PR #54, opened 2019-02-21
support multiple protocols

soenkehahn:
Maybe easier to review commit by commit.


soenkehahn:
Fixes #53.

src/emulation/mod.rs
```diff
@@ -186,3 +191,16 @@ mod run_against_protocol {
...
+pub fn run_against_protocols(
+    context: Context,
+    executable: &Path,
+    expected: Vec<Protocol>,
+) -> R<TestResults> {
+    Ok(TestResults(
+        expected
+            .into_iter()
+            .map(|expected| run_against_protocol(context.clone(), executable, expected))
+            .collect::<R<Vec<TestResult>>>()?,
```
astampoulis: Is the explicit type instantiation required here?

src/emulation/mod.rs
```diff
@@ -94,8 +96,11 @@ impl SyscallMock {
     }
 
     fn push_error(&mut self, error: String) {
-        if self.errors.is_none() {
-            self.errors = Some(error);
+        match self.errors {
```
astampoulis: From what I see `TestResult` can only store a single error. How come you use the plural `errors`? The name together with the fact that this method is called `push_error` confused me a bit when reading this method, as I was expecting this to add an error to some list of errors. However, if there is an error already, the new error is ignored.

src/emulation/mod.rs
soenkehahn: Yes. I suspect because of the `?`. The `?` implicitly converts errors (inside `Result::Err`) to different errors (using `From`). So -- without the turbofish -- the result of `collect` would be allowed to be any type `Result<Vec<TestResult>, E>` where `E` can be converted to our error type (`Box<std::error::Error>`, see [here](https://github.com/Originate/check-protocols/blob/master/src/lib.rs#L15)). The turbofish pins down the error type.

It could be shortened to `::<R<_>>`. Do you think that'd be better?

src/emulation/mod.rs
soenkehahn: Good point. I've changed the names. Is this clearer now?

astampoulis: Yeah, it reads better to me!

src/emulation/mod.rs
astampoulis: I see, that makes sense. I would keep the full type in that case!

Issue #72, opened 2019-02-22
No way to mock file/folder existence 


matthandlersux:
i have a ruby script that checks for files/folders (ex: `if File.exists?("/usr/local/bin/unrar")`. i don't believe that this call is intercepted by check-protocols, and i'm not sure if it's possible to intercept it. this would likely not be a problem if i could run the tool on my machine, but since it is running within docker, the files/folders don't exist.

if it's not possible to intercept these types of calls, i would ideally be able to create a folder tree as part of my protocol. off the top of my head, something like the following could work:
```yaml
filetree:
  usr: 
    local:
      bin: unrar
protocol:
  ...
```
or
```yaml
filetree:
  - /usr/local/bin/unrar
protocol:
  ...
```

in my particular case, i check for the existence of executables, directories, and regular files.

Issue #67, opened 2019-02-21
ability to test commands that include arguments with spaces


astampoulis:
For example, ability to specify a protocol like:
```yaml
- executable: /usr/bin/g it
  arguments:
    - commit
    - -m
    - re encrypted
```
for:
`"g it" commit -m "re ecnrypted"`

PR #73, opened 2019-02-22
add script for local development on macs (in docker)

### LINKED ISSUE #74: allow to contribute to the codebase on OSX

matthandlersux:
i went to start adding a feature (#72) and realized that local development on OSX wasn't very smooth. this PR adds [cargo-watch](https://github.com/passcod/cargo-watch) to the docker image and creates a script to run the tests continuously within docker.

Closes #74.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM rust:1.32.0
 
+RUN cargo install cargo-watch
+
 WORKDIR /root/check-protocols
 ADD Cargo.* ./
 RUN mkdir src && touch src/lib.rs && cargo install --root /usr/local --path . ; true
--- a/README.md
+++ b/README.md
@@ -126,5 +126,6 @@ run inside of docker. Luckily, we have set up a one-liner for you. This will run
 within docker, when files change:
 
 ``` bash
+./build-docker-image.sh
 ./test-watch-in-docker.sh
 ```
--- a/README.md
+++ b/README.md
@@ -102,11 +102,29 @@ protocol:
   - git pull
 ```
 
-## Running inside `docker`
+## Running inside `docker` (for OSX)
 
 You can run the tool inside docker, for example like this:
 
 ``` bash
 ./build-docker-image.sh
 ./check-protocols-in-docker.sh <PATH_TO_YOUR_SCRIPT>
+```
+
+## Contributing
+
+Contributions, feature requests, bug reports, etc. are all welcome. Please consider the following guidelines
+when submitting:
+
+* For any pull request that you intend to merge, we ask that all tests pass for every commit that will end up on master
+* We will address the top rated (by :thumbsup:) issues first, please cast your votes!
+
+### For OSX
+
+This tool does not currently compile or run on OSX. In order to develop on a Mac you will need to
+run inside of docker. Luckily, we have set up a one-liner for you. This will run the tests continuously,
+within docker, when files change:
+
+``` bash
+./test-watch-in-docker.sh
 ```
--- a/test-watch-in-docker.sh
+++ b/test-watch-in-docker.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+docker run --init --rm -it --cap-add=SYS_PTRACE \
+  --entrypoint "/bin/bash" -v $(pwd):/root/ \
+  check-protocols -c "cargo watch -x 'test -- --test-threads=1 --quiet'"


test-watch-in-docker.sh.protocols.yaml
```diff
@@ -0,0 +1 @@
+- /usr/bin/docker run --init --rm -it --cap-add=SYS_PTRACE --entrypoint /bin/bash -v /root/project:/root/ check-protocols -c cargo watch -x test
```
matthandlersux: the tests fail because of the way quotes are handled between script <-> protocol. i'm not positive that this isn't my own fault, however... running locally:

```sh
# script (with quotes included)
docker run --init --rm -it --cap-add=SYS_PTRACE --entrypoint "/bin/bash" -v $(pwd):/root/ check-protocols -c "cargo watch -x test"
# protocol (with quotes included at the end)
- /usr/bin/docker run --init --rm -it --cap-add=SYS_PTRACE --entrypoint /bin/bash -v /root:/root/ check-protocols -c "cargo watch -x test"
```
i get
```
error:
  expected: /usr/bin/docker run --init --rm -it --cap-add=SYS_PTRACE --entrypoint /bin/bash -v /root:/root/ check-protocols -c "cargo watch -x test"
  received: /usr/bin/docker run --init --rm -it --cap-add=SYS_PTRACE --entrypoint /bin/bash -v /root:/root/ check-protocols -c cargo watch -x test
```
`received` makes me think that i should remove the quotes in my protocol, but if i do that, i get
```
error:
  expected: /usr/bin/docker run --init --rm -it --cap-add=SYS_PTRACE --entrypoint /bin/bash -v /root:/root/ check-protocols -c cargo watch -x test
  received: /usr/bin/docker run --init --rm -it --cap-add=SYS_PTRACE --entrypoint /bin/bash -v /root:/root/ check-protocols -c cargo watch -x test
```
(they are the same string, i double checked that there were no hidden characters)

...i dug into this, it looks like the `arguments` here https://github.com/Originate/check-protocols/blob/master/src/protocol/mod.rs#L63 don't match (possibly something that should be added to the formatted error output?)

```
# self.arguments
["run", "--init", "--rm", "-it", "--cap-add=SYS_PTRACE", "--entrypoint", "/bin/bash", "-v", "/root:/root/", "check-protocols", "-c", "cargo", "watch", "-x", "test"]
# arguments
["run", "--init", "--rm", "-it", "--cap-add=SYS_PTRACE", "--entrypoint", "/bin/bash", "-v", "/root:/root/", "check-protocols", "-c", "cargo watch -x test"]
```

soenkehahn: Yeah, quoted arguments don't really work right now. There's this ticket for that: #67.

soenkehahn: I think this shouldn't block this PR though. This PR would also be good without the test protocol. As much as I like using the tool itself in this repo, this protocol is also not *that* useful, since it just repeats the command from the script.

matthandlersux: agreed, they aren't very helpful for one liners. i'll remove this.

README.md
```diff
@@ -102,11 +102,17 @@ protocol:
...
 ``` bash
 ./build-docker-image.sh
 ./check-protocols-in-docker.sh <PATH_TO_YOUR_SCRIPT>
 ```
+
+and for local development, you can run tests continuously
+
+``` bash
+./test-watch-in-docker.sh
+```
```
soenkehahn: Now there are two fairly different use-cases for the docker image:

- Run the tool as a user.
- Hack on the rust code of the tool.

I think that's fine, but I think this should be separated more in the README. Maybe a section about running the tool in docker and then another section `Contributing` that would include how to develop on a mac? If you want I can take a stab at that.

Dockerfile
```diff
@@ -6,6 +6,7 @@ RUN mkdir src && touch src/lib.rs && cargo install --root /usr/local --path . ;
 ADD src ./src
 RUN touch src/lib.rs
 RUN cargo install --root /usr/local --path .
+RUN cargo install cargo-watch
```
soenkehahn: This `Dockerfile` is optimized for running fast when only some source code module of `check-protocols` changes. For example line 5 installs only the dependencies, so that that is cached separately from compiling the code in this repo. I would put this command at the top of the file (after `FROM ...`).

soenkehahn:
Closes #74.

README.md
matthandlersux: updated

README.md
```diff
@@ -102,11 +102,29 @@ protocol:
...
+when submitting:
+
+* For any pull request that you intend to merge, we ask that all tests pass for every commit that will end up on master
+* We will address the top rated (by :thumbsup:) issues first, please cast your votes!
+
+### For OSX
+
+This tool does not currently compile or run on OSX. In order to develop on a Mac you will need to
+run inside of docker. Luckily, we have set up a one-liner for you. This will run the tests continuously,
+within docker, when files change:
```
soenkehahn: Maybe this should mention that you have to run `./build-docker-image.sh` first.

PR #79, opened 2019-02-25
output better error messages when yaml parsing fails

### LINKED ISSUE #77: indicate clearly when there's a yaml parse error
Currently an error gets reported, but it's not always clear that the error comes from trying to parse the yaml file. Example:

`error: block sequence entries are not allowed in this context`

From this yaml file:
``` yaml
protocol: - foo
```

It should say something like:

`error parsing foo.protocols.yaml: block sequence entries are not allowed in this context`

@matthandlersux: I think you mentioned this somewhere, but I can't find it.

soenkehahn:
Fixes #77.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -222,10 +222,21 @@ impl Protocol {
         })
     }
 
-    pub fn load(executable_path: &Path) -> R<Vec<Protocol>> {
-        let file_contents = read_protocols_file(executable_path)?;
+    fn parse_protocols(file_contents: &str) -> R<Vec<Protocol>> {
         let yaml: Vec<Yaml> = YamlLoader::load_from_str(&file_contents)?;
         yaml.into_iter().map(Protocol::parse).collect()
+    }
+
+    pub fn load(executable_path: &Path) -> R<Vec<Protocol>> {
+        let file_contents = read_protocols_file(executable_path)?;
+        let result = Protocol::parse_protocols(&file_contents).map_err(|error| {
+            format!(
+                "error parsing {}: {}",
+                add_extension(executable_path, "protocols.yaml").to_string_lossy(),
+                error
+            )
+        })?;
+        Ok(result)
     }
 }
 
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -222,20 +222,27 @@ impl Protocol {
         })
     }
 
-    fn parse_protocols(file_contents: &str) -> R<Vec<Protocol>> {
-        let yaml: Vec<Yaml> = YamlLoader::load_from_str(&file_contents)?;
-        yaml.into_iter().map(Protocol::parse).collect()
-    }
-
     pub fn load(executable_path: &Path) -> R<Vec<Protocol>> {
-        let file_contents = read_protocols_file(executable_path)?;
-        let result = Protocol::parse_protocols(&file_contents).map_err(|error| {
+        let protocols_file = find_protocol_file(executable_path);
+        let file_contents = read_protocols_file(&protocols_file)?;
+        let yaml: Vec<Yaml> = YamlLoader::load_from_str(&file_contents).map_err(|error| {
             format!(
-                "error parsing {}: {}",
-                add_extension(executable_path, "protocols.yaml").to_string_lossy(),
+                "invalid YAML in {}: {}",
+                protocols_file.to_string_lossy(),
                 error
             )
         })?;
+        let result = yaml
+            .into_iter()
+            .map(Protocol::parse)
+            .collect::<R<Vec<Protocol>>>()
+            .map_err(|error| {
+                format!(
+                    "unexpected type in {}: {}",
+                    protocols_file.to_string_lossy(),
+                    error
+                )
+            })?;;
         Ok(result)
     }
 }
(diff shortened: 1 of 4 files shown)


src/protocol/mod.rs
```diff
@@ -222,11 +222,22 @@ impl Protocol {
...
-    pub fn load(executable_path: &Path) -> R<Vec<Protocol>> {
-        let file_contents = read_protocols_file(executable_path)?;
+    fn parse_protocols(file_contents: &str) -> R<Vec<Protocol>> {
         let yaml: Vec<Yaml> = YamlLoader::load_from_str(&file_contents)?;
         yaml.into_iter().map(Protocol::parse).collect()
     }
+
+    pub fn load(executable_path: &Path) -> R<Vec<Protocol>> {
+        let file_contents = read_protocols_file(executable_path)?;
+        let result = Protocol::parse_protocols(&file_contents).map_err(|error| {
```
matthandlersux: is it necessary to unwrap the result with `?`, assign it, and then rewrap with `Ok(result)` below?

tests/run.rs
```diff
@@ -45,6 +50,50 @@ fn simple() -> R<()> {
...
+    fn wrong_types() -> R<()> {
+        let script = TempFile::write_temp_script("")?;
+        let result = test_run_check_protocols(
+            &script,
+            r##"
+                |protocol: 42
+            "##,
+        );
+        assert_eq!(
+            format!("{}", result.unwrap_err()),
```
matthandlersux: just curious, is this equivalent to `result.unwrap_err().to_string()`?

test-utils/src/lib.rs
```diff
@@ -23,6 +23,7 @@ fn run(command: &str, args: Vec<&str>) -> R<()> {
     }
 }
 
+#[derive(Debug)]
```
matthandlersux: i don't think you make use of this in this PR

src/protocol/mod.rs
```diff
@@ -222,11 +222,22 @@ impl Protocol {
...
+            format!(
+                "error parsing {}: {}",
```
matthandlersux: i think if the message mentions YAML specifically it would be more clear, something like: `YAML parsing error in: {}`

tests/run.rs
```diff
@@ -45,6 +50,50 @@ fn simple() -> R<()> {
...
+            format!(
+                "error parsing {}.protocols.yaml: expected: array, got: Integer(42)",
```
matthandlersux: if i'm reading it correctly, this is not a YAML mistake, it's a misuse of the library. it would be nice if the error message could clarify between the two, so that one tells me (essentially) "you made a YAML mistake" and the other tells me "you misread the `check-protocols` docs"

src/protocol/mod.rs
soenkehahn: Yeah, it is. `map_err` returns a value of type `Result<Vec<Protocol>, String>`. The result of `load` is `R<Vec<Protocol>>` which expands to `Result<Vec<Protocol>, Box<std::error::Error>>`. The `?` implicitly converts the error with the `From` trait. So we need that.

What would work is this:

``` rust
Ok(Protocol::parse_protocols(&file_contents).map_err(|error| {
    format!(
        "error parsing {}: {}",
        add_extension(executable_path, "protocols.yaml").to_string_lossy(),
        error
    )
})?)
```

Would that be better?

src/protocol/mod.rs
soenkehahn: The file name will always include the `.yaml` extension, that's why I thought it wouldn't be necessary. Do you still think this change would be useful?

test-utils/src/lib.rs
soenkehahn: Yes, removed.

tests/run.rs
soenkehahn: Yes, it is. TIL. `format("{}", ...)` uses the `Display` trait, but `ToString` is automatically implemented for everything that implements `Display`. (https://doc.rust-lang.org/std/string/trait.ToString.html#tymethod.to_string)

tests/run.rs
soenkehahn: Yeah, that's correct. I changed the error messages now, Are they better like that?

src/protocol/mod.rs
```diff
@@ -224,8 +224,25 @@ impl Protocol {
...
+        let result = yaml
+            .into_iter()
+            .map(Protocol::parse)
+            .collect::<R<Vec<Protocol>>>()
+            .map_err(|error| {
+                format!(
+                    "unexpected type in {}: {}",
+                    add_extension(executable_path, "protocols.yaml").to_string_lossy(),
+                    error
+                )
```
soenkehahn: I want to still dry up the two `map_err` calls.

src/protocol/mod.rs
matthandlersux: ah, implicits. hmm, i don't think that would be any better, i think this is always bound to happen when implicit stuff occurs (where some lines of code aren't immediately obvious). in my mind the ideal thing to see would be a more explicit conversion, like...

```rust
Protocol::parse_protocols(&file_contents).map_err(|error| {
    format!(
        "error parsing {}: {}",
        add_extension(executable_path, "protocols.yaml").to_string_lossy(),
        error
    ).to_boxed() // here
})
```
but that could just be my lack of experience with rust

tests/run.rs
matthandlersux: yup 👍

PR #82, opened 2019-02-26
trace commands that are executed using vfork

### LINKED ISSUE #81: allow to track commands in ruby scripts
Currently in ruby scripts like the one below, no commands are being detected.

``` ruby
#!/usr/bin/env ruby
`touch foo`
```

soenkehahn:
This is how some languages, e.g. ruby, execute subprocesses.

Fixes #81.


astampoulis:
:+1:

PR #100, opened 2019-03-01
add error message for failing execve

### LINKED ISSUE #69: output a good error message when the hashbang interpreter of the tested script can't be found
Example:
``` bash
#!/does/not/exist
something something
```

data-pup:
Fixes #69.

### DIFF
--- a/src/emulation/mod.rs
+++ b/src/emulation/mod.rs
@@ -144,7 +144,7 @@ mod run_against_protocol {
     use super::*;
     use crate::protocol::command::Command;
     use std::fs;
-    use test_utils::{assert_error, trim_margin, TempFile};
+    use test_utils::{trim_margin, TempFile};
 
     #[test]
     fn works_for_longer_file_names() -> R<()> {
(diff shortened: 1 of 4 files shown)


src/emulation/mod.rs
```diff
@@ -178,18 +178,6 @@ mod run_against_protocol {
         Ok(())
     }
 
-    #[test]
-    fn complains_when_the_file_does_not_exist() {
```
data-pup: as mentioned while pairing, this is not applicable now

src/syscall_mocking/mod.rs
```diff
@@ -40,6 +42,34 @@ impl Tracer {
         }
     }
 
+    fn execve(
```
data-pup: because nicer error messages increase complexity of the closure given to `fork_with_child_errors`, this has been outlined.

tests/run.rs
```diff
@@ -156,6 +147,54 @@ fn failing_later() -> R<()> {
     Ok(())
 }
 
+mod nice_user_errors {
```
data-pup: tests :tada:

PR #103, opened 2019-03-02
include more elaborate example in readme

### LINKED ISSUE #98: Add more script examples to documentation
Once more of this project stabilises, we might want to set up documentation with [mdbook](https://github.com/rust-lang-nursery/mdBook). This would let us publish more detailed docs to GitHub Pages :)

One example would be talking about protocols in more depth. The example in the `README` covers a simple shell script, but the largest benefit of this idea comes when scripts in other languages shell out with dynamically generated commands. We could cover a simple example like that, and so on.

soenkehahn:
This is a stab at #98. I think something like this is a good start. @data-pup: Should we close #98 or keep it open?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,25 +29,40 @@ Feel free to open (or vote on)
 `check-protocols` allows you to write tests -- so-called protocols -- for
 scripts (or other executables) -- without the need to modify your executables.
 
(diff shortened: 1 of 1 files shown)


data-pup:
@soenkehahn I think #98 can be closed once this is merged

PR #109, opened 2019-03-04
add fcntl debugging

soenkehahn:


src/syscall_mocking/debugging.rs
```diff
@@ -116,3 +120,26 @@ impl Debugger {
...
+#[allow(non_camel_case_types)]
+enum FcntlCmd {
+    F_DUPFD,
+    F_GETFD,
+    F_SETFD,
+    F_GETFL,
+    F_SETFL,
+}
+
+impl From<c_ulonglong> for FcntlCmd {
```
data-pup: nice! :tada:

src/syscall_mocking/debugging.rs
```diff
@@ -116,3 +120,26 @@ impl Debugger {
...
+    fn from(fcntl_cmd: c_ulonglong) -> Self {
+        match fcntl_cmd {
+            0 => FcntlCmd::F_DUPFD,
+            1 => FcntlCmd::F_GETFD,
+            2 => FcntlCmd::F_SETFD,
+            3 => FcntlCmd::F_GETFL,
+            4 => FcntlCmd::F_SETFL,
+            _ => panic!("FcntlCmd: unknown cmd: {}", fcntl_cmd),
```
data-pup: out of curiousity, why `panic!` here instead of throwing an `Error` of some sort?

soenkehahn: Good point. I thought it wouldn't matter, since this shouldn't happen. But in the rest of this `debugging` module I stick to the policy that debugging should never fail, neither with a panic nor with an error. I'll change it.

PR #110, opened 2019-03-04
Km/script args with spaces

### LINKED ISSUE #66: allow to specify arguments with spaces that get passed to the tested script
`check-protocols` allows to pass in arguments to the tested script:

``` bash
#!/usr/bin/env bash
mv $1 $2
```

``` yaml
arguments: foo bar
protocol:
  - mv foo bar
```

The string passed in through the `arguments` field is split into arguments by spaces. So currently there is no way to pass in arguments that contain spaces. (The arguments parsing should follow the same rules as the parsing for expected arguments for the individual steps, as in #67.)

data-pup:
Fixes #66.

### DIFF
--- a/src/protocol/argument_parser.rs
+++ b/src/protocol/argument_parser.rs
@@ -1,0 +1,118 @@
+use crate::R;
+use std::collections::VecDeque;
+
+#[derive(Debug)]
+pub(super) struct Parser {
+    pub original: String,
+    pub input: VecDeque<char>,
+}
+
+impl Parser {
+    fn parse_error<A>(&self, message: &str) -> R<A> {
+        Err(format!("{} ({:?})", message, self.original))?
+    }
+
+    fn skip_spaces(&mut self) {
+        loop {
+            match self.input.pop_front() {
+                Some(' ') => {}
+                Some(char) => {
+                    self.input.push_front(char);
+                    break;
+                }
+                None => break,
+            }
+        }
+    }
+
+    fn skip_char(&mut self, expected: char, message: &str) -> R<()> {
+        let next = self.input.pop_front();
+        if next != Some(expected) {
+            self.parse_error(message)?;
+        }
+        Ok(())
+    }
+
+    fn peek_chars(&mut self, expected: Vec<Option<char>>, message: &str) -> R<()> {
+        let peeked = self.input.get(0);
+        if !expected.contains(&peeked.cloned()) {
+            self.parse_error(message)?;
+        }
+        Ok(())
+    }
+
+    fn parse_char(&mut self, excluded: &[char]) -> R<Option<char>> {
+        Ok(match self.input.pop_front() {
+            None => None,
+            Some('\\') => Some(match self.input.pop_front() {
+                None => self.parse_error("a backslash must be followed by a character")?,
+                Some('"') => '"',
+                Some('n') => '\n',
+                Some(' ') => ' ',
+                Some('\\') => '\\',
+                Some(char) => self.parse_error(&format!("unknown escaped character {}", char))?,
+            }),
+            Some(char) => {
+                if excluded.contains(&char) {
+                    self.input.push_front(char);
+                    None
+                } else {
+                    Some(char)
+                }
+            }
+        })
+    }
+
+    fn collect_chars_until(&mut self, excluded: &[char]) -> R<String> {
+        let mut result = "".to_string();
+        while let Some(char) = self.parse_char(excluded)? {
+            result.push(char);
+        }
+        Ok(result)
+    }
+
+    fn parse_word(&mut self) -> R<Option<String>> {
+        self.skip_spaces();
+        Ok(match self.input.get(0) {
+            None => None,
+            Some('"') => {
+                self.skip_char('"', "shouldn't happen")?;
+                let word = self.collect_chars_until(&['"'])?;
+                self.skip_char('"', "unmatched quotes")?;
+                self.peek_chars(
+                    vec![Some(' '), None],
+                    "closing quotes must be followed by a space",
+                )?;
+                Some(word)
+            }
+            Some(_) => {
+                let result = self.collect_chars_until(&[' ', '"'])?;
+                self.peek_chars(
+                    vec![Some(' '), None],
+                    "opening quotes must be preceeded by a space",
+                )?;
+                Some(result)
+            }
+        })
+    }
+
+    pub fn parse_arguments(arguments: &str) -> R<Vec<String>> {
+        Self {
+            original: arguments.to_owned(),
+            input: arguments.chars().collect(),
+        }
+        .collect()
+    }
+}
+
+impl Iterator for Parser {
+    type Item = R<String>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.parse_word() {
+            Ok(Some(word)) => Some(Ok(word)),
+            Ok(None) => None,
+            Err(error) => Some(Err(error)),
+        }
+    }
+}
--- a/src/protocol/command.rs
+++ b/src/protocol/command.rs
@@ -1,3 +1,4 @@
+use super::argument_parser::Parser;
 use crate::R;
 
 #[derive(PartialEq, Eq, Debug, Clone)]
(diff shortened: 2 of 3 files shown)


src/protocol/command.rs
```diff
@@ -42,124 +43,6 @@ impl Command {
     }
 
     pub fn new(string: &str) -> R<Command> {
-        use std::collections::VecDeque;
```
data-pup: we start by outlining the `Parser` struct and placing it into `src/protocol/parser.rs`.

src/protocol/mod.rs
```diff
@@ -194,12 +195,15 @@ impl Protocol {
     }
 
     fn add_arguments(&mut self, object: &Hash) -> R<()> {
-        if let Ok(arguments) = object.expect_field("arguments") {
-            self.arguments = arguments
-                .expect_str()?
-                .split_whitespace()
-                .map(String::from)
-                .collect();
+        if let Ok(arguments) = object
```
data-pup: here is where we call the new logic for parsing script arguments when creating a `Protocol`

src/protocol/mod.rs
```diff
@@ -550,6 +538,59 @@ mod load {
         Ok(())
     }
 
+    mod script_arguments {
```
data-pup: tests for script arguments live here

src/protocol/parser.rs
```diff
@@ -0,0 +1,130 @@
+use crate::R;
+use std::collections::VecDeque;
+use super::Command;
+
+#[derive(Debug)]
+pub(super) struct Parser {
```
data-pup: Here is where we move the `Parser` struct to. This is `pub(super)` because it should only be visible to the `src/protocol/mod.rs` and `src/protocol/command.rs` files.

soenkehahn:
Fixes #66.

src/protocol/parser.rs
```diff
@@ -0,0 +1,130 @@
...
+                Some(word) => arguments.push(word.into_bytes()),
+            }
+        }
+        Ok(Command {
+            executable,
+            arguments,
+        })
+    }
+
+    pub fn parse_script_arguments(&mut self) -> R<Vec<String>> {
```
data-pup: here is the new method for parsing script arguments, following the same style as above in `parse_command`

src/protocol/parser.rs
```diff
@@ -99,32 +99,36 @@ impl Parser {
         })
     }
 
-    pub fn parse_command(&mut self) -> R<Command> {
-        let executable = match self.parse_word()? {
-            None => self.parse_error("expected: space-separated command and arguments")?,
+    pub fn parse_command(command: &str) -> R<Command> {
```
data-pup: here's where things get opinionated-ish. This is in a separate commit so we can remove these changes if they aren't desired.

The gist of these is that we adjust `parse_command` and `parse_script_arguments` from methods that can mutate `self` into public static methods that receive a single argument. This is helpful because the callers in `protocol/mod.rs` and `protocol/command.rs` no longer need to worry about cloning input into the `original` and `input` fields.

The other stylistic change is that the `match` nested within a `loop` is refactored as a `while let Some(_)` loop, when parsing arguments into a vector of inputs.

src/protocol/mod.rs
```diff
@@ -199,11 +200,7 @@ impl Protocol {
             .expect_field("arguments")
             .and_then(|arguments| arguments.expect_str())
         {
-            self.arguments = parser::Parser {
-                original: arguments.to_string(),
-                input: arguments.chars().collect()
-            }
-            .parse_script_arguments()?
+            self.arguments = Parser::parse_script_arguments(arguments)?
```
data-pup: here's an example of how the changes to the `parse_*` methods affect the call sites. I think this takes some of the complexity out of these calls, since we no longer have to worry about the interior structure of the parsing struct, and can simply pass a `&str` to receive our result.

data-pup:
This _might_ be easier to review commit by commit, I divided out the feature work, and some minor alterations into separate commits. :)

src/protocol/parser.rs
soenkehahn: I wonder if it would be good to move the creation of `Command` out into where this is called. So that this module only deals with argument parsing and nothing else. Then we'd have a nice module with a tiny interface that's just `parse_arguments(string: &str) -> Vec<String>`.

src/protocol/parser.rs
```diff
@@ -0,0 +1,130 @@
+use crate::R;
```
soenkehahn: I would call this module `argument_parser`. Or something that distinguishes this from the yaml parsing in `protocol`.

src/protocol/mod.rs
```diff
@@ -194,12 +195,15 @@ impl Protocol {
...
     fn add_arguments(&mut self, object: &Hash) -> R<()> {
-        if let Ok(arguments) = object.expect_field("arguments") {
-            self.arguments = arguments
-                .expect_str()?
-                .split_whitespace()
-                .map(String::from)
-                .collect();
+        if let Ok(arguments) = object
+            .expect_field("arguments")
+            .and_then(|arguments| arguments.expect_str())
```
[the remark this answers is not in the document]
data-pup: good call. added some tests for that case. because we don't want to throw an error when there are no arguments, this part can be a little tricky!

src/protocol/argument_parser.rs
```diff
@@ -0,0 +1,118 @@
...
+                self.peek_chars(
+                    vec![Some(' '), None],
+                    "opening quotes must be preceeded by a space",
+                )?;
+                Some(result)
+            }
+        })
+    }
+
+    pub fn parse_arguments(arguments: &str) -> R<Vec<String>> {
```
data-pup: The only public function now is `parse_arguments`, which returns a `Result<Vec<String>, _>` :tada: 

This is mostly a pleasant abstraction over the `Iterator` implementation below, so that we don't have to use cumbersome turbofish calls to `collect`

src/protocol/command.rs
```diff
@@ -41,130 +42,24 @@ impl Command {
...
+                executable: executable.clone().into_bytes(),
+                arguments: arguments
+                    .iter()
+                    .map(|arg| arg.clone().into_bytes())
+                    .collect(),
+            })
+            .ok_or_else(|| {
+                format!(
+                    "{} ({:?})",
+                    "expected: space-separated command and arguments",
```
soenkehahn: This can be put into the format string itself, right?

src/protocol/command.rs
```diff
@@ -41,130 +42,24 @@ impl Command {
...
-        }
-
-        Ok(Parser {
-            original: string.to_string(),
-            input: string.chars().collect(),
-        }
-        .parse_command()?)
+    pub fn new(command: &str) -> R<Command> {
+        Parser::parse_arguments(command)?
+            .split_first()
```
soenkehahn: `split_first` is nice. I always wanted a replacement for pattern matching on list cons'es.

src/protocol/command.rs
```diff
@@ -41,130 +42,24 @@ impl Command {
...
+            .map(|(executable, arguments)| Command {
+                executable: executable.clone().into_bytes(),
+                arguments: arguments
+                    .iter()
+                    .map(|arg| arg.clone().into_bytes())
```
soenkehahn: Could you use `into_iter` and then get rid of the `clone`?

src/protocol/command.rs
```diff
@@ -41,130 +42,24 @@ impl Command {
...
+                    .collect(),
+            })
+            .ok_or_else(|| {
+                format!(
+                    "{} ({:?})",
+                    "expected: space-separated command and arguments",
+                    command.to_string()
+                )
+                .into()
+            })
```
soenkehahn: Maybe this would look better as a `match  Parser::parse_arguments(command)?.split_first() { ... }`.

src/protocol/mod.rs
```diff
@@ -550,6 +532,64 @@ mod load {
...
+                    "##
+                ),
+                format!(
+                    "unexpected type in {}.protocols.yaml: \
+                     expected: string, got: Integer(42)",
+                    path_to_string(&tempfile.path())?
+                )
+            );
+            Ok(())
+        }
```
soenkehahn: :+1:

PR #112, opened 2019-03-04
WIP mocking file existence

### LINKED ISSUE #72: No way to mock file/folder existence

soenkehahn:
closes #72

### DIFF
--- a/src/emulation/mod.rs
+++ b/src/emulation/mod.rs
@@ -76,6 +76,18 @@ impl SyscallMock {
                     ptrace::setregs(pid, registers)?;
                 }
             }
+            (Syscall::Stat, SyscallStop::Exit) if !self.protocol.mocked_files.is_empty() => {
+                let filename = tracee_memory::peek_string(pid, registers.rdi)?;
+                if self.protocol.mocked_files.contains(&filename) {
+                    if filename.ends_with(b"/") {
+                        let buffer_ptr = registers.rsi;
+                        tracee_memory::poke_four_bytes(pid, buffer_ptr + 24, libc::S_IFDIR as i32)?;
+                    }
+                    let mut registers = *registers;
+                    registers.rax = 0;
+                    ptrace::setregs(pid, registers)?;
+                }
+            }
             _ => {}
         }
         Ok(())
@@ -140,14 +152,14 @@ impl SyscallMock {
 pub fn run_against_protocol(
     context: Context,
     executable: &Path,
-    expected: Protocol,
+    protocol: Protocol,
     unmocked_commands: &[Vec<u8>],
 ) -> R<TestResult> {
     let syscall_mock = Tracer::run_against_mock(
         executable,
-        expected.arguments.clone(),
-        expected.env.clone(),
-        |tracee_pid| SyscallMock::new(context, tracee_pid, expected, unmocked_commands),
+        protocol.arguments.clone(),
+        protocol.env.clone(),
+        |tracee_pid| SyscallMock::new(context, tracee_pid, protocol, unmocked_commands),
     )?;
     Ok(syscall_mock.result)
 }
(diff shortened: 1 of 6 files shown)


src/emulation/mod.rs
```diff
@@ -78,14 +78,12 @@ impl SyscallMock {
...
-                    "mocked files: {:?} - {:?}",
-                    self.protocol.mocked_files.clone().into_iter().map(|f| String::from_utf8(f).unwrap()).collect::<Vec<String>>(),
-                    String::from_utf8(filename.clone())
-                );
                 if self.protocol.mocked_files.contains(&filename) {
-                    println!("huhu");
-                    let mut registers = registers.clone();
+                    if filename.ends_with(b"/") {
+                        let buffer_ptr = registers.rsi;
+                        tracee_memory::pokedata(pid, buffer_ptr + 24, u64::from(libc::S_IFDIR))?;
```
matthandlersux: @soenkehahn i figured out how to find the location in memory where the file mode is specified in `*statbuf` 😄 (this was a stretch for me... some relevant links [statbuf struct def](https://doc.servo.org/libc/unix/notbsd/linux/other/b64/x86_64/struct.stat.html), [descriptions of the fields](http://man7.org/linux/man-pages/man2/fstat.2.html), [extended info about the fields](http://man7.org/linux/man-pages/man7/inode.7.html))

...but i'm going to need some advice on how to clean this up... i'm not in love with the magic number, but unwrapping the entire struct isn't really necessary. also `pokedata` and `ptrace::write` write 8 bytes but `S_IFDIR` is actually a `u32` (so i assume that means i'm clobbering the `st_uid` field, which could have side effects).

src/emulation/mod.rs
```diff
@@ -78,14 +78,12 @@ impl SyscallMock {
...
                 let filename = tracee_memory::peek_string(pid, registers.rdi)?;
-                println!(
-                    "mocked files: {:?} - {:?}",
-                    self.protocol.mocked_files.clone().into_iter().map(|f| String::from_utf8(f).unwrap()).collect::<Vec<String>>(),
-                    String::from_utf8(filename.clone())
-                );
                 if self.protocol.mocked_files.contains(&filename) {
-                    println!("huhu");
-                    let mut registers = registers.clone();
+                    if filename.ends_with(b"/") {
```
matthandlersux: @soenkehahn question: do you think it's ok to start with this as the way to specify to `check-protocols` that you intend for a file path in `foldertree` to be treated as a directory? 

the problem i imagine is, i may accidentally try to mock a directory `/foo/bar/` like this:
```yaml
filetree:
  - /foo/bar
  - /foo/bar/test
```
but if the tests fail, it will be hard to figure out why.

on the other hand it could just be mentioned in the readme, hopefully in a clear enough way that people won't run into this by accident.

matthandlersux:
@soenkehahn i'm going to close this for now, since it's still WIP, but should be done shortly.

one question i have for you... should we also automatically mock the existence of commands that are in our protocol? ie, if i have `command: /bin/something`, should `if [ -f /bin/something ]; then` succeed?

soenkehahn:
There is a separate issue for this: #114.

PR #113, opened 2019-03-04
add ruby to dockerfile

matthandlersux:
the tests currently fail within docker (using `test-watch-in-docker.sh`) because the tests depend on the existence of ruby. this fixes that.

@soenkehahn would it make sense for circle to run the tests within the same container that is used for local development?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM rust:1.32.0
 
 RUN cargo install cargo-watch
+RUN apt-get update && apt-get install -y ruby
 
 WORKDIR /root/check-protocols
 ADD Cargo.* ./


Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM rust:1.32.0
 
 RUN cargo install cargo-watch
 
+RUN apt-get update
+RUN apt-get install -y ruby-full
```
soenkehahn: This should be in one line with a `&&`. There's an explanation for why here: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

soenkehahn: Do we need `ruby-full`? Not that I'm that worried about image size, but we're only executing the tiniest ruby script.

Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM rust:1.32.0
 
 RUN cargo install cargo-watch
```
soenkehahn: I would have one less newline. To have all the dependencies in one stanza.

Dockerfile
matthandlersux: no, we don't, i didn't realize there was a `ruby` package

PR #115, opened 2019-03-05
update debugger

matthandlersux:
this PR updates the syscall debugger to 
* display the open syscall with a filename
* log syscall details...
    * before calls that are handled on enter
    * after calls that are handled on exit

### DIFF
--- a/src/emulation/mod.rs
+++ b/src/emulation/mod.rs
@@ -44,9 +44,9 @@ impl SyscallMock {
     pub fn handle_syscall(
         &mut self,
         pid: Pid,
-        syscall_stop: SyscallStop,
-        syscall: Syscall,
-        registers: user_regs_struct,
+        syscall_stop: &SyscallStop,
+        syscall: &Syscall,
+        registers: &user_regs_struct,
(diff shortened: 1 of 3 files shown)


matthandlersux:
@soenkehahn clippy suggests:
```
error: using `clone` on a `Copy` type
  --> src/emulation/mod.rs:74:41
   |
74 |                     let mut registers = registers.clone();
   |                                         ^^^^^^^^^^^^^^^^^ help: try dereferencing it: `*registers`
```
...which i can do, but i'm not sure i understand why. maybe it will make more sense to me in the morning.

soenkehahn:
For types that implement `Copy` the compiler will implicitly add calls to `clone` whenever they're necessary. So here dereferencing will just be a `clone`. Didn't know `user_regs_struct` implemented `Copy`.

matthandlersux:
@soenkehahn thanks for the explanation. in your opinion, which one is clearer? i still prefer the `registers.clone()` just because it's explicit.

soenkehahn:
@matthandlersux: I don't really have that much of an opinion.

PR #116, opened 2019-03-05
don't generate example tests file

soenkehahn:
Having a build.rs file that generates a test file
makes cargo rebuild lots of things unnecessarily.
This change speeds up the development feedback
loop.


soenkehahn:
Yeah... is there a better way to get rid of files that a tool in the repo used to generate?

soenkehahn:
@amnda-d had the idea to create multiple commits to
- check the generated file into the repo,
- `git rm` it.
That could work...

matthandlersux:
i don't know a way :'(

PR #118, opened 2019-03-05
move integration test utils into their own module

soenkehahn:


tests/run.rs
```diff
@@ -1,30 +1,18 @@
+#![cfg_attr(
+    feature = "dev",
```
matthandlersux: i think this means that this file is only compiled when `--cfg feature="dev"` is passed to rustc... but what i'm not sure about is what you intend to consider "dev". is that for running tests locally? does it get run on CI? curious if there's a more descriptive feature name than "dev" that could make this clearer.

tests/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+    allow(dead_code, unused_variables, unused_imports, unreachable_code)
+)]
+#![deny(clippy::all)]
+
+use check_protocols::{run_check_protocols, Context, ExitCode, R};
+use pretty_assertions::assert_eq;
+use std::fs;
+use test_utils::{trim_margin, TempFile};
+
+pub fn test_run_check_protocols(script: &TempFile, protocol: &str) -> R<(ExitCode, String)> {
```
matthandlersux: this function is only used 3 times in `tests/run.rs`, could it be made private in favor of only calling `test_run` there? i'm mostly wondering because i think the naming between the two functions isn't very clear.

it looks like you're using it so that you can get access to the script path since it appears in the error message you assert on, so maybe this could be called `test_run_with_tempfile`?

tests/run.rs
soenkehahn: No, this doesn't exclude the file from compilation. This says that if `--features dev` is passed to `cargo` that a few warnings are disabled. I use this during development to suppress certain warnings that I usually don't want to see while working on the code.

tests/utils.rs
soenkehahn: Good point. Comparing with `test_run` it also doesn't call `compare_results`. I'll rename it to what you said.

tests/run.rs
matthandlersux: ah i see, the feature flag applies to the args below. makes more sense then 👍

PR #120, opened 2019-03-06
add yaml interpreter field

### LINKED ISSUE #89: allow to specify an interpreter for a script file
Currently `check-protocols` can only be run on executables. That means the tested files have to:

- have the executable flag set,
- have a hashbang at the top (or be elf binaries).

It should be possible to get around these restrictions by providing an interpreter, for example like this:

``` yaml
interpreter: python3
tests:
    - arguments: asdff asdfdsakl 
      protocol: 
        - asdfhhhhh
        - aljflkds

    - arguments: fjfjfjf
      protocol:
        - thing1
        - thing2

    - arguments: ofheoheff
      protocol:
        - thing1
```

The current workaround for this is to just make the scripts executable and put the hashbang in, but we don't want to force our users to run their scripts in a certain way.

cc: @xasopheno

data-pup:
fixes #89. :smiley_cat:

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,11 @@ protocols:
         exitcode?: number
           # Mocked exitcode of the command.
           # Default: 0
+interpreter?: string
+    # The interpreter that should be used to run the tested script.
+    # Example: "/bin/bash", default: The program itself will be executed
+    # directly, without an interpreter. In that case it has to have the
+    # executable flag set. Often you also will need a hashbang.
 unmockedCommands: [string]
   # List of executables that are not going to be mocked out, but are going to be
   # executed instead.
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -249,6 +249,7 @@ impl Protocol {
 pub struct Protocols {
     pub protocols: Vec<Protocol>,
     pub unmocked_commands: Vec<Vec<u8>>,
+    pub interpreter: Option<Vec<u8>>,
 }
 
 impl Protocols {
@@ -260,7 +261,15 @@ impl Protocols {
         Ok(Protocols {
             protocols: result,
             unmocked_commands: vec![],
+            interpreter: None,
         })
+    }
+
+    fn add_interpreter(&mut self, object: &Hash) -> R<()> {
+        if let Ok(interpreter) = object.expect_field("interpreter") {
+            self.interpreter = Some(interpreter.expect_str()?.as_bytes().to_vec());
+        }
+        Ok(())
     }
 
     fn add_unmocked_commands(&mut self, object: &Hash) -> R<()> {
@@ -283,11 +292,13 @@ impl Protocols {
                 (Ok(protocols), _) => {
                     let mut protocols = Protocols::from_array(protocols.expect_array()?)?;
                     protocols.add_unmocked_commands(object)?;
+                    protocols.add_interpreter(object)?;
                     protocols
                 }
                 (Err(_), Ok(_)) => Protocols {
                     protocols: vec![Protocol::from_object(&object)?],
                     unmocked_commands: vec![],
+                    interpreter: None,
                 },
                 (Err(_), Err(_)) => Err(format!(
                     "expected field \"protocol\" or \"protocols\", got: {:?}",
@@ -698,6 +709,51 @@ mod load {
                 .unmocked_commands
                 .map(|command| String::from_utf8(command).unwrap()),
                 vec!["foo"]
+            );
+            Ok(())
+        }
+    }
+
+    mod specified_interpreter {
+        use super::*;
+        use pretty_assertions::assert_eq;
+
+        #[test]
+        fn allows_a_specified_interpreter() -> R<()> {
+            let tempfile = TempFile::new()?;
+            assert_eq!(
+                test_parse(
+                    &tempfile,
+                    r##"
+                        |protocols:
+                        |  - protocol: []
+                        |interpreter: /bin/bash
+                    "##,
+                )?
+                .interpreter
+                .unwrap(),
+                b"/bin/bash",
+            );
+            Ok(())
+        }
+
+        #[test]
+        fn disallows_an_interpreter_with_an_invalid_type() -> R<()> {
+            let tempfile = TempFile::new()?;
+            assert_error!(
+                test_parse(
+                    &tempfile,
+                    r##"
+                        |protocols:
+                        |  - protocol: []
+                        |interpreter: 42
+                    "##,
+                ),
+                format!(
+                    "unexpected type in {}.protocols.yaml: \
+                     expected: string, got: Integer(42)",
+                    path_to_string(&tempfile.path())?
+                )
             );
             Ok(())
         }
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -10,8 +10,8 @@ pub fn path_to_string(path: &Path) -> R<&str> {
         .ok_or_else(|| format!("invalid utf8 sequence: {:?}", &path))?)
 }
 
-pub fn parse_shebang(executable: &Path) -> Option<String> {
-    let contents = fs::read(executable).ok()?;
+pub fn parse_shebang(program: &Path) -> Option<String> {
+    let contents = fs::read(program).ok()?;
     if contents.starts_with(b"#!") {
         let bytes = contents
             .into_iter()
--- a/tests/run.rs
+++ b/tests/run.rs
@@ -30,6 +30,23 @@ fn simple() -> R<()> {
     Ok(())
 }
 
+#[test]
+fn can_specify_interpreter() -> R<()> {
+    test_run(
+        r##"
+            |`true`
+        "##,
+        r##"
+            |protocols:
+            |  - protocol:
+            |    - /bin/true
+            |interpreter: /usr/bin/ruby
+        "##,
+        Ok(()),
+    )?;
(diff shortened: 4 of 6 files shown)


README.md
```diff
@@ -98,6 +98,11 @@ protocols:
         exitcode?: number
           # Mocked exitcode of the command.
           # Default: 0
+interpreter: string
+    # The interpreter that should be used to run the tested script. If no
+    # interpreter is specified, a `#!` line at the top of the script will be
+    # used to identify the interpreter.
+    # Example: "/bin/bash", default: ""
```
data-pup: here's where the new interpreter field is documented in `README`. lmk if there are any edits I should make here!

src/emulation/executable_mock.rs
```diff
@@ -15,7 +15,7 @@ pub fn create_mock_executable(context: &Context, config: Config) -> R<Vec<u8>> {
     let mut result = b"#!".to_vec();
     result.append(
         &mut context
-            .check_protocols_executable
+            .check_protocols_program
```
data-pup: because we might not be testing the behaviour of an executable if the interpreter is specified, this has been renamed.

src/emulation/mod.rs
```diff
@@ -137,18 +139,47 @@ impl SyscallMock {
     }
 }
 
+pub enum Executable {
+    Interpreter(PathBuf),
+    Program(PathBuf),
+}
```
data-pup: This enum exists to differentiate between an executable that refers to the interpreter that should be invoked, i.e. `ruby`, and a program that is itself executable.

src/emulation/mod.rs
```diff
@@ -137,18 +139,47 @@ impl SyscallMock {
...
+
+impl Deref for Executable {
+    type Target = Path;
+    fn deref(&self) -> &Path {
+        match self {
+            Executable::Interpreter(program) => &program,
+            Executable::Program(program) => &program,
+        }
```
data-pup: We implement `Deref` on the `Executable` enum so that we can coerce it into a `&Path` when calling `execve`

src/emulation/mod.rs
```diff
@@ -137,18 +139,47 @@ impl SyscallMock {
...
     expected: Protocol,
     unmocked_commands: &[Vec<u8>],
 ) -> R<TestResult> {
-    let syscall_mock = Tracer::run_against_mock(
-        executable,
-        expected.arguments.clone(),
-        expected.env.clone(),
-        |tracee_pid| SyscallMock::new(context, tracee_pid, expected, unmocked_commands),
-    )?;
+    let (executable, arguments): (Executable, Vec<String>) = match interpreter {
```
data-pup: initialize the `executable` and `arguments` values depending on whether we are running a script directly, or invoking it using an interpreter

src/protocol/mod.rs
```diff
@@ -249,6 +249,7 @@ impl Protocol {
 pub struct Protocols {
     pub protocols: Vec<Protocol>,
     pub unmocked_commands: Vec<Vec<u8>>,
+    pub interpreter: Option<Vec<u8>>,
```
data-pup: This is the new field added to `Protocols` used to represent the specified interpreter, if given

src/protocol/mod.rs
```diff
@@ -283,11 +292,13 @@ impl Protocols {
                 (Ok(protocols), _) => {
                     let mut protocols = Protocols::from_array(protocols.expect_array()?)?;
                     protocols.add_unmocked_commands(object)?;
+                    protocols.add_interpreter(object)?;
```
data-pup: If we are given a `protocols:` object in the yaml file, here is where we add the interpreter. The function is defined above :)

src/protocol/mod.rs
```diff
@@ -702,6 +713,51 @@ mod load {
...
+                        |interpreter: 42
+                    "##,
+                ),
+                format!(
+                    "unexpected type in {}.protocols.yaml: \
+                     expected: string, got: Integer(42)",
+                    path_to_string(&tempfile.path())?
+                )
+            );
+            Ok(())
```
data-pup: Tests to check that the interpreter parsed, and that an error is thrown if the value is not of the correct type, are defined here

src/syscall_mocking/mod.rs
```diff
@@ -45,33 +45,42 @@ impl Tracer {
...
-            };
-            format!(
-                "execve'ing {} failed with error: {}\n{}",
-                executable.to_string_lossy(),
-                error,
-                hint,
-            )
-        })?;
+        execve(&path, &c_args, &c_env)
+            .map_err(|error| Self::format_execve_error(executable, &args, error))?;
```
data-pup: The closure for formatting an error given by `execve` has been outlined into a function above.

src/syscall_mocking/mod.rs
```diff
@@ -45,33 +45,42 @@ impl Tracer {
...
+            }
+        };
+        format!(
+            "execve'ing {} failed with error: {}\n{}",
+            program, error, hint,
+        )
+    }
+
+    fn execve(executable: &Executable, args: Vec<String>, env: HashMap<String, String>) -> R<()> {
         let path = CString::new(executable.as_os_str().as_bytes())?;
```
data-pup: This is the line that depends on our `Deref` implementation for `Executable`.

src/emulation/mod.rs
```diff
@@ -137,18 +139,47 @@ impl SyscallMock {
...
+        Some(interpreter) => {
+            let executable = String::from_utf8_lossy(interpreter.as_ref())
+                .into_owned()
+                .into();
+            let mut arguments = vec![path_to_string(program).unwrap().to_string()];
```
data-pup: FIXUP: use https://doc.rust-lang.org/std/path/struct.Path.html#method.to_string_lossy

thx @matthandlersux

src/emulation/mod.rs
data-pup: This comment is now out of date, found a better way to do this :)

src/emulation/mod.rs
```diff
@@ -137,18 +139,57 @@ impl SyscallMock {
...
+}
+
+impl Into<Vec<u8>> for &Executable {
+    fn into(self) -> Vec<u8> {
+        let path = match self {
+            Executable::Interpreter(path) => path,
+            Executable::Program(path) => path,
+        };
+        path.as_os_str().as_bytes().to_vec()
+    }
```
data-pup: We implement `Into<Vec<u8>>` for Executable, because the [CString constructor](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new) works on anything that implements `Into<Vec<u8>>`. I have some slightly mixed feelings about the fact that this is implemented on a reference, but changing this would mean cloning Executable afaict, which seems worse

src/syscall_mocking/mod.rs
```diff
@@ -45,33 +47,73 @@ impl Tracer {
...
+                c_args.push_front(CString::new(interpreter.clone())?);
+                (
+                    CString::new(interpreter)?,
+                    c_args.into_iter().collect(),
+                    c_env,
+                )
+            }
+            None => (c_executable.clone(), c_args.into_iter().collect(), c_env),
+        })
+    }
```
soenkehahn: Yeah, this is actually pretty good now, I think. Pushing the interpreter name to the front of that argument list intuitively makes the most sense to me.

src/syscall_mocking/mod.rs
```diff
@@ -45,33 +47,73 @@ impl Tracer {
...
-        })?;
+        };
+        let hint = format!("Does {} exist?", interpreter);
+        format!(
+            "execve'ing {} failed with error: {}\n{}",
+            program.to_string_lossy(),
+            error,
+            hint,
+        )
+    }
```
soenkehahn: Nice.

src/syscall_mocking/mod.rs
```diff
@@ -45,33 +47,73 @@ impl Tracer {
...
+
+    fn execve(
+        interpreter: Option<Vec<u8>>,
```
soenkehahn: Maybe this can be passed in by reference? And then only cloned in the case of errors? Not sure whether that would make things better.

src/utils/mod.rs
```diff
@@ -10,8 +10,9 @@ pub fn path_to_string(path: &Path) -> R<&str> {
         .ok_or_else(|| format!("invalid utf8 sequence: {:?}", &path))?)
 }
 
-pub fn parse_shebang(executable: &Path) -> Option<String> {
-    let contents = fs::read(executable).ok()?;
+pub fn parse_shebang(program: &Path) -> Option<String> {
+    let path = Path::new(program);
+    let contents = fs::read(path).ok()?;
```
soenkehahn: Why was this changed?

tests/run.rs
```diff
@@ -177,7 +194,77 @@ mod nice_user_errors {
...
+                format!(
+                    r##"
+                        |execve'ing {} failed with error: ENOEXEC: Exec format error
+                        |Does your interpreter exist?
+                    "##,
+                    path_to_string(script.path().as_ref())?
+                )
+                .as_str(),
+            )?
+            .trim()
```
soenkehahn: Why are we `trim`ming this here? Does the `result` not include a trailing `\n`? If not, shouldn't it?

soenkehahn: Errors don't have a trailing newline.

PR #121, opened 2019-03-06
step line numbers are recorded when protocol is loaded (#65)

### LINKED ISSUE #65: when test fails, indicate which step failed more clearly
If I have this protocol:
```
- foo
- bar
- foo
```
And this script:
```
foo
bar
baz
```
I would like the error reporting to look something like this:
```
- foo
- bar
vvvvvvvv
- foo
^^^^^^^
error happened here
```
(I just mean something that points to the exact line, e.g. a number or something like that.)

hallettj:
Adds a failing test that asserts that the line number for each protocol step is recorded when loading a protocol. fixes #65

@soenkehahn

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ path = "src/main.rs"
 nix = "*"
 libc = "*"
 tempdir = "*"
-yaml-rust = "*"
+yaml-rust = { git = "https://github.com/hallettj/yaml-rust", branch = "load_from_str_with_markers" }
 linked-hash-map = "*"
 bincode = "*"
 quale = "*"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ path = "src/main.rs"
 nix = "*"
 libc = "*"
 tempdir = "*"
-yaml-rust = { git = "https://github.com/hallettj/yaml-rust", branch = "load_from_str_with_markers" }
+yaml-rust = { git = "https://github.com/hallettj/yaml-rust", rev = "5216abe1f7d24ccd17a3b1b1e56c0114758d5ed7" }
 linked-hash-map = "*"
 bincode = "*"
 quale = "*"
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -104,6 +104,7 @@ impl Step {
 
 #[cfg(test)]
 mod parse_step {
+    use self::test_helpers::step_with_marker;
     use super::*;
     use pretty_assertions::assert_eq;
     use test_utils::assert_error;
@@ -114,12 +115,6 @@ mod parse_step {
         assert_eq!(yaml.len(), 1);
         let yaml = &yaml[0];
         Step::parse(yaml)
-    }
-
-    fn step_with_marker(command_matcher: CommandMatcher, marker: Marker) -> Step {
-        let mut step = Step::new(command_matcher);
-        step.marker = Some(marker);
-        step
     }
 
     #[test]
@@ -520,6 +515,7 @@ impl Protocols {
 
 #[cfg(test)]
 mod load {
+    use self::test_helpers::step_with_marker;
     use super::*;
     use crate::R;
     use pretty_assertions::assert_eq;
@@ -538,12 +534,6 @@ mod load {
         let result = test_parse(&tempfile, protocol_string)?.protocols;
         assert_eq!(result.len(), 1);
         Ok(result.into_iter().next().unwrap())
-    }
-
-    fn step_with_marker(command_matcher: CommandMatcher, marker: Marker) -> Step {
-        let mut step = Step::new(command_matcher);
-        step.marker = Some(marker);
-        step
     }
 
     #[test]
@@ -1327,5 +1317,16 @@ mod find_protocol_file {
             find_protocol_file(&PathBuf::from("foo.ext")),
             PathBuf::from("foo.ext.protocols.yaml")
         );
+    }
+}
+
+#[cfg(test)]
+mod test_helpers {
+    use super::*;
+
+    pub fn step_with_marker(command_matcher: CommandMatcher, marker: Marker) -> Step {
+        let mut step = Step::new(command_matcher);
+        step.marker = Some(marker);
+        step
     }
 }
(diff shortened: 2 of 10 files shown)


matthandlersux:
i hadn't heard about [draft pull requests](https://github.blog/2019-02-14-introducing-draft-pull-requests/) yet :)

hallettj:
I have an upstream pull request with yaml-rust here:

https://github.com/chyh1990/yaml-rust/pull/125

I have updated the branch in this PR to use my fork of yaml-rust, report line numbers in error messages, and adjust tests as necessary to match the new output.

Cargo.toml
```diff
@@ -15,7 +15,7 @@ path = "src/main.rs"
 nix = "*"
 libc = "*"
 tempdir = "*"
-yaml-rust = "*"
+yaml-rust = { git = "https://github.com/hallettj/yaml-rust", branch = "load_from_str_with_markers" }
```
soenkehahn: I think we should pin this to a commit that is guaranteed to compile. (I think you can do that with a `rev` field.)

src/protocol/mod.rs
```diff
@@ -31,33 +34,39 @@ impl Step {
...
             self.stdout = stdout.expect_str()?.as_bytes().to_vec();
         }
         Ok(())
     }
 
-    fn parse(yaml: &Yaml) -> R<Step> {
-        match yaml {
-            Yaml::String(string) => Step::from_string(string),
-            Yaml::Hash(object) => {
+    fn parse(Node(yaml, marker): &Node) -> R<Step> {
```
matthandlersux: ooh, i didn't know you could destructure in function signatures

src/protocol/mod.rs
```diff
@@ -94,24 +105,34 @@ impl Step {
...
-            Step::new(CommandMatcher::ExactMatch(Command {
-                executable: PathBuf::from("foo"),
-                arguments: vec![],
-            })),
+            step_with_marker(
+                CommandMatcher::ExactMatch(Command {
+                    executable: PathBuf::from("foo"),
+                    arguments: vec![],
+                }),
+                Marker { line: 1, col: 0 }
```
matthandlersux: since the marker is a default value, what do you think about making the function above `step_with_default_marker` and not specifying the marker explicitly in the tests?

src/protocol/mod.rs
```diff
@@ -514,6 +540,12 @@ mod load {
         Ok(result.into_iter().next().unwrap())
     }
 
+    fn step_with_marker(command_matcher: CommandMatcher, marker: Marker) -> Step {
```
matthandlersux: i assume you had to duplicate the function above because here it is in a different sub `mod`? i ran into this same issue at one point and wasn't sure the best way to share the function between both. seems like you could either add a larger surrounding `mod` (and move the function up a level), or import the function from the other one like `use super::mod_name::function_name`. not sure what the best rusty way to do it is, curious what you and @soenkehahn think though.

src/protocol/yaml.rs
```diff
@@ -3,32 +3,40 @@ use linked_hash_map::LinkedHashMap;
...
+    fn expect_object(&self) -> R<&LinkedHashMap<Self::Child, Self::Child>>;
 
     fn expect_integer(&self) -> R<i32>;
 }
 
-impl YamlExt for Yaml {
+impl<T> YamlExt for T
+where
+    T: fmt::Debug + YamlNode,
+    <T as YamlNode>::Child: fmt::Debug,
```
matthandlersux: can you help me understand how to interpret this? 

is `<T as YamlNode>::Child` just the way to specify the generic `Child` inside of `T`? then, since we specify on the line above that `T: YamlNode`, this allows us to add bounds to the `Child`?

is there a name for this pattern that i could look up?

src/protocol/mod.rs
hallettj: Yep! But there is a weird exception where (AFAIK) you cannot destructure an argument if it is `self` because there is just no syntax for that.

Cargo.toml
hallettj: Ok, I have made that change. The revision is locked in `Cargo.lock` - but I do see the benefit of giving a revision in `Cargo.toml` to make sure that nothing breaks when someone runs `cargo update`.

src/protocol/mod.rs
hallettj: I was kinda torn on this. The tests make an equality comparison between `Step` values, and if the marker is not specified correctly in the expected value then the test will fail. We could change the tests to compare specific properties within a `Step` instead of the whole structure - for example we could compare just the command and arguments. In that case calling `Step::new` would suffice instead of `step_with_marker`.

Honestly I'm trying to suppress my instinct to put multiple assertions in a test (to check equality for command and arguments) in favor of the idea of testing just one thing in each test. I suppose I could put the commands and arguments into tuples for comparisons to technically get one assertion. I might make that change.

src/protocol/mod.rs
hallettj: I did not think of importing from a sibling module. Maybe I could move `step_with_marker` into a test helper module. I try not to worry too much about duplicating code in tests. But there is no reason for duplication if it is easy to avoid.

src/protocol/yaml.rs
hallettj: Yes, you've got it. This is a constraint on an associated type. The methods in `YamlExt` use `Debug` implementations to return errors, which means that both `T` and the `Child` associated type must implement `Debug`. And yes, `<T as YamlNode>::Child` is a way to reference the associated type in the `YamlNode` impl for whatever type `T` happens to be. I don't know the specific term for casting a type variable to a trait to access its associated types, I only know how to describe it in those terms.

Another way to make this work would have been to put the `Debug` constraint on the associated type in the `YamlExt` trait, instead of in the impl like this:

```rust
trait YamlExt {
    type Child: fmt::Debug;
    // ...
}
```

But it is in principle possible to implement `YamlExt` with a `Child` type that does not implement `Debug`; and my instruction in Haskell was to keep trait / type class bounds as close as possible to implementations that actually require them to keep reusable code as general as is reasonable.

matthandlersux: cool, thanks for the explanation!

PR #124, opened 2019-03-07
resolve full path of executables in path steps

### LINKED ISSUE #32: Allow to specify protocols with executables from the $PATH
Right now you have to say e.g.
```
- /usr/bin/git ...
```
But it should be allowed to say
```
- git ...
```

hallettj:
fixes #32


hallettj:
@soenkehahn I've made the necessary changes get path resolution working, and to get tests passing on my machine. It looks like `/bin/true` and `/bin/false` don't map to files in `$PATH` on CircleCI, so I still need to resolve that issue. It would be helpful to get a review of the changes that I have made.

src/syscall_mock/mod.rs
```diff
@@ -20,7 +19,7 @@ pub struct SyscallMock {
     context: Context,
     tracee_pid: Pid,
     protocol: Protocol,
-    unmocked_commands: Vec<Vec<u8>>,
+    unmocked_commands: Vec<PathBuf>,
```
hallettj: I could change this type from `PathBuf` back to `Vec<u8>`. That would remove the need to make `Comand::path_from_bytes` a public method. On the other hand `PathBuf` or `Path` seem like the "correct" type for an OS command.

src/syscall_mock/mod.rs
```diff
@@ -158,3 +167,47 @@ impl SyscallMock {
...
+            syscall_mock.result,
+            TestResult::Failure(trim_margin(
+                r#"
+                    |  expected: cp
+                    |  received: mv
+                "#
+            )?)
+        );
+        Ok(())
+    }
```
soenkehahn: I think I would prefer to test this with an integration test in `tests/path.rs`. I hope that would

1. look a bit nicer and simpler,
2. not rely on implementation details as much.

src/syscall_mock/mod.rs
```diff
@@ -158,3 +167,47 @@ impl SyscallMock {
...
+        let unmocked_commands = vec![Vec::from("cp")];
+        let syscall_mock = SyscallMock::new(
+            &Context::new()?,
+            Pid::from_raw(123),
+            Protocol::empty(),
+            unmocked_commands.as_slice(),
+        );
+        assert!(syscall_mock.unmocked_commands[0].is_absolute());
+        Ok(())
+    }
```
soenkehahn: I don't think we should test that `SyscallMock::new` performs this translation. That's an implementation detail that we ideally would want to be able to refactor without having the tests fail. Or was there a special corner-case or other reason to add this test?

src/syscall_mock/mod.rs
```diff
@@ -109,11 +114,15 @@ impl SyscallMock {
...
     }
 
-    fn handle_step(&mut self, received: protocol::Command) -> R<PathBuf> {
+    fn handle_step(&mut self, received: Command) -> R<PathBuf> {
         let mock_config = match self.protocol.steps.pop_front() {
             Some(next_protocol_step) => {
-                if next_protocol_step.command != received {
-                    self.register_error(&next_protocol_step.command.format(), &received.format());
+                let expected_command = next_protocol_step.command.resolve_executable_path();
+                if expected_command != received {
```
soenkehahn: Maybe we should move this into a `Command.compare` function? So that the resolution of executable paths happens inside that function?

src/syscall_mock/mod.rs
```diff
@@ -36,7 +35,10 @@ impl SyscallMock {
             context: context.clone(),
             tracee_pid,
             protocol,
-            unmocked_commands: unmocked_commands.to_vec(),
+            unmocked_commands: unmocked_commands
+                .iter()
+                .map(|c| Command::executable_full_path(c))
+                .collect(),
```
soenkehahn: I'm not sure it's a good idea to change the behaviour of `unmockedCommands` in this PR. Just to keep things small. Also, is this change tested?

src/protocol/command.rs
```diff
@@ -224,6 +282,86 @@ mod command {
                 );
                 Ok(())
             }
+
```
soenkehahn: Remove this newline?

src/protocol/command.rs
```diff
@@ -224,6 +282,86 @@ mod command {
...
+            fn updates_command_with_shortened_exeucutable_if_in_path() -> R<()> {
+                let command = Command::new("cp")?;
+                assert_eq!(
+                    command
+                        .resolve_executable_path()
+                        .shorten_executable_in_path(),
+                    command
+                );
+                Ok(())
+            }
```
soenkehahn: I would try to separate the tests that test `resolve_executable_path` from the ones that test `executable_file_name`. Into their own submodules.

tests/run.rs
```diff
@@ -880,16 +880,16 @@ mod unmocked_commands {
             r##"
                 |protocols:
                 |  - protocol:
-                |    - /usr/bin/dirname dir/file
-                |    - /bin/ls dir
+                |    - dirname dir/file
+                |    - ls dir
```
data-pup: Haven't reviewed some of the other changes in this PR fully quite yet, but I might leave these existing tests as is? That way if there is a regression regarding path resolution at some point, it is easier to pinpoint why tests are failing

src/syscall_mock/mod.rs
hallettj: I changed the type of `unmocked_commands` back to `Vec<Vec<u8>>`.

src/syscall_mock/mod.rs
hallettj: Ok, I replaced these tests with integration tests.

src/syscall_mock/mod.rs
hallettj: Good point; this is now tested with an integration test, `looks_up_unmocked_command_executable_in_path`.

src/syscall_mock/mod.rs
hallettj: Sounds like a good idea to me. I implemented `compare` for checking received commands against protocol steps, and `compare_executables` for checking commands against `unmocked_executables`. I was not sure whether it would be preferable stylistically to implement `compare` as `pub fn compare(&self, other: &Command)` or `pub fn compare(a: &Command, b: &Command)`. I went with the former for now.

src/syscall_mock/mod.rs
hallettj: Ok, I changed it back, and implemented `compare_executables` to achieve the same result which keeps path resolution logic in `protocol::command`. But the new implementation does perform more `$PATH` lookups.

src/protocol/command.rs
hallettj: done

src/protocol/command.rs
hallettj: Ok; I moved tests into a separate module for each method under test.

tests/run.rs
hallettj: Unfortunately the tests fail on my machine if I do not remove the paths from most of the integration test protocol steps.

tests/path.rs
```diff
@@ -0,0 +1,69 @@
...
+    )?;
+    Ok(())
+}
+
+#[test]
+fn looks_up_unmocked_command_executable_in_path() -> R<()> {
+    test_run(
+        r##"
+            |#!/usr/bin/env bash
+            |ls $(dirname dir/file)
```
soenkehahn: This line can be simplified to just `dirname dir/file`. (In general I'm trying to keep the tests as simple as possible and only test one thing at a time.)

tests/path.rs
```diff
@@ -0,0 +1,69 @@
...
+        Err(&trim_margin(
+            "
+                |error:
+                |  expected: cp
+                |  received: mv
+            ",
+        )?),
+    )?;
+    Ok(())
+}
```
soenkehahn: Maybe we should have a test for what happens when the script executes e.g. `/usr/bin/something`, and the protocol specifies `something`, but the first thing found in the `$PATH` is `/usr/local/bin/something`.

Similarly we probably want a test that makes sure the user *can* specify that a script is supposed to execute `/usr/bin/something`, even though it's shadowed by `/usr/local/bin/something`.

And maybe we also need a test that make sure users can still specify executables that are not in the `$PATH` at all.

src/protocol/command.rs
```diff
@@ -28,8 +44,32 @@ impl Command {
         word.chars().map(escape_char).collect::<Vec<_>>().join("")
     }
 
+    fn executable_with_path(executable: &[u8]) -> PathBuf {
```
soenkehahn: Can we rename this to something with `canonicalize` in the name?

src/protocol/command.rs
```diff
@@ -28,8 +44,32 @@ impl Command {
...
+                let resolved = Command::executable_with_path(file_name.as_bytes());
+                if resolved == path {
+                    file_name.to_os_string()
+                } else {
+                    path.into_os_string()
+                }
+            }
+            None => path.as_os_str().to_os_string(),
+        }
+    }
```
soenkehahn: What do you think about pulling out the functions `compare_executables`, `executable_with_path` and `executable_short_form` from `Command`? They could be standalone functions in this module (or in a new module, maybe?).

(And then the tests for these functions could be paired with the functions, i.e. the matching test module could be directly after each function.)

src/protocol/command.rs
```diff
@@ -28,8 +44,16 @@ impl Command {
         word.chars().map(escape_char).collect::<Vec<_>>().join("")
     }
 
+    /// Resolve the path of an executable by searching $PATH. Falls back to the original executable
+    /// value if no result is found in $PATH.
+    fn executable_full_path(executable: &[u8]) -> PathBuf {
```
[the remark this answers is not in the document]
hallettj: I changed the names for expanding and shortening the path to `canonicalize_executable_path` and `canonicalize_executable_short_form`.

tests/path.rs
hallettj: Sounds good. I simplified this test to `ls > /dev/null` to get a single command that will not crash the script, and that will not produce output.

tests/path.rs
hallettj: I added tests for a script that runs an executable that is not the first match in `$PATH` where the protocol expects an executable that is the first match, for a script that runs the first match in path where the protocol expects an executable that is not the first match, and where the script runs an executable that is not in `$PATH` at all.

src/protocol/command.rs
hallettj: As mentioned in my other comment, I renamed this function to `canonicalize_executable_path`.

src/protocol/command.rs
hallettj: Sounds like a good plan. I moved the relevant functions and unit tests to `src/protocol/path_expansion.rs`.

PR #126, opened 2019-03-08
filetree mocking

### LINKED ISSUE #72: No way to mock file/folder existence

matthandlersux:
reopening a clean version of #112. i believe this is ready for review now. the only open question i am aware of is whether `filetree` is what we want to call it. 

~~also there appears to be some nondeterministic error with this on CI that i can't reproduce locally. i'll try to hunt this down tomorrow.~~ ...now that i can write to the `statbuf`, i can do it for a regular file too, which i believe fixed the issue on CI (and it means that we no longer need to edit the `Dockerfile`)

closes #72

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,10 @@ protocols:
         exitcode?: number
           # Mocked exitcode of the command.
           # Default: 0
+        mockedFiles?: [string]
+          # List of files and folders that are going to be mocked to exist.
+          # Note that directories must include a trailing '/'.
+          # Example: ["/www/logs"], default: []
 interpreter?: string
     # The interpreter that should be used to run the tested script.
     # Example: "/bin/bash", default: The program itself will be executed
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -171,6 +171,7 @@ pub struct Protocol {
     pub env: HashMap<String, String>,
     pub cwd: Option<Vec<u8>>,
     pub exitcode: i32,
+    pub mocked_files: Vec<Vec<u8>>,
 }
 
 impl Protocol {
@@ -186,6 +187,7 @@ impl Protocol {
             env: HashMap::new(),
             cwd: None,
             exitcode: 0,
+            mocked_files: vec![],
         }
     }
 
@@ -235,12 +237,23 @@ impl Protocol {
         Ok(())
     }
 
+    fn add_mocked_files(&mut self, object: &Hash) -> R<()> {
+        if let Ok(paths) = object.expect_field("mockedFiles") {
+            for path in paths.expect_array()?.iter() {
+                self.mocked_files
+                    .push(path.expect_str()?.as_bytes().to_owned());
+            }
+        }
+        Ok(())
+    }
+
     fn from_object(object: &Hash) -> R<Protocol> {
         let mut protocol = Protocol::from_array(object.expect_field("protocol")?.expect_array()?)?;
         protocol.add_arguments(&object)?;
         protocol.add_env(&object)?;
         protocol.add_cwd(&object)?;
         protocol.add_exitcode(&object)?;
+        protocol.add_mocked_files(&object)?;
         Ok(protocol)
     }
 }
@@ -757,6 +770,23 @@ mod load {
             );
             Ok(())
         }
+    }
+
+    #[test]
+    fn allows_to_specify_mocked_files() -> R<()> {
+        assert_eq!(
+            test_parse_one(
+                r##"
+                    |protocol: []
+                    |mockedFiles:
+                    |  - /foo
+                "##
+            )?
+            .mocked_files
+            .map(|path| String::from_utf8_lossy(&path).into_owned()),
+            vec![("/foo")]
+        );
+        Ok(())
     }
 }
 
--- a/src/syscall_mock/mod.rs
+++ b/src/syscall_mock/mod.rs
@@ -76,6 +76,26 @@ impl SyscallMock {
                     ptrace::setregs(pid, registers)?;
                 }
             }
+            (Syscall::Stat, SyscallStop::Exit) => {
+                let filename = tracee_memory::peek_string(pid, registers.rdi)?;
+                if self.protocol.mocked_files.contains(&filename) {
+                    let statbuf_ptr = registers.rsi;
+                    let mock_mode = if filename.ends_with(b"/") {
+                        libc::S_IFDIR
+                    } else {
+                        libc::S_IFREG
+                    };
+                    #[allow(clippy::forget_copy)]
+                    tracee_memory::poke_four_bytes(
+                        pid,
+                        statbuf_ptr + (offset_of!(libc::stat, st_mode) as u64),
+                        mock_mode as u32,
+                    )?;
+                    let mut registers = *registers;
+                    registers.rax = 0;
+                    ptrace::setregs(pid, registers)?;
+                }
+            }
             _ => {}
         }
         Ok(())
@@ -141,15 +161,15 @@ pub fn run_against_protocol(
     context: Context,
     interpreter: &Option<Vec<u8>>,
     program: &Path,
-    expected: Protocol,
+    protocol: Protocol,
     unmocked_commands: &[Vec<u8>],
 ) -> R<TestResult> {
     let syscall_mock = Tracer::run_against_mock(
         interpreter,
         program,
-        expected.arguments.clone(),
-        expected.env.clone(),
-        |tracee_pid| SyscallMock::new(context, tracee_pid, expected, unmocked_commands),
+        protocol.arguments.clone(),
+        protocol.env.clone(),
+        |tracee_pid| SyscallMock::new(context, tracee_pid, protocol, unmocked_commands),
     )?;
     Ok(syscall_mock.result)
 }
--- a/src/tracer/tracee_memory.rs
+++ b/src/tracer/tracee_memory.rs
@@ -1,22 +1,39 @@
 use crate::R;
-use libc::{c_ulonglong, c_void};
+use libc::{c_uint, c_ulonglong, c_void};
 use nix::sys::ptrace;
 use nix::unistd::Pid;
 
-fn cast_to_byte_array(word: c_ulonglong) -> [u8; 8] {
-    let ptr: &[u8; 8];
-    unsafe {
-        ptr = &*(&word as *const u64 as *const [u8; 8]);
-    }
-    *ptr
+fn cast_to_four_byte_array(x: c_uint) -> [u8; 4] {
+    [
+        (x & 0xff) as u8,
+        ((x >> 8) & 0xff) as u8,
+        ((x >> 16) & 0xff) as u8,
+        ((x >> 24) & 0xff) as u8,
+    ]
 }
 
-fn cast_to_word(bytes: [u8; 8]) -> c_ulonglong {
-    let void_ptr;
-    unsafe {
-        void_ptr = std::mem::transmute(bytes);
-    }
-    void_ptr
+fn cast_to_eight_byte_array(x: c_ulonglong) -> [u8; 8] {
+    [
+        (x & 0xff) as u8,
+        ((x >> 8) & 0xff) as u8,
+        ((x >> 16) & 0xff) as u8,
+        ((x >> 24) & 0xff) as u8,
+        ((x >> 32) & 0xff) as u8,
+        ((x >> 40) & 0xff) as u8,
+        ((x >> 48) & 0xff) as u8,
+        ((x >> 56) & 0xff) as u8,
+    ]
+}
+
+fn cast_to_eight_byte_word(bytes: [u8; 8]) -> c_ulonglong {
+    u64::from(bytes[0])
+        + (u64::from(bytes[1]) << 8)
+        + (u64::from(bytes[2]) << 16)
+        + (u64::from(bytes[3]) << 24)
+        + (u64::from(bytes[4]) << 32)
+        + (u64::from(bytes[5]) << 40)
+        + (u64::from(bytes[6]) << 48)
+        + (u64::from(bytes[7]) << 56)
 }
 
 fn peekdata(pid: Pid, address: c_ulonglong) -> R<c_ulonglong> {
@@ -45,7 +62,7 @@ fn peekdata_iter(pid: Pid, address: c_ulonglong) -> impl Iterator<Item = R<c_ulo
 fn data_to_string(data: impl Iterator<Item = R<c_ulonglong>>) -> R<Vec<u8>> {
     let mut result = vec![];
     'outer: for word in data {
-        for char in cast_to_byte_array(word?).iter() {
+        for char in cast_to_eight_byte_array(word?).iter() {
             if *char == 0 {
                 break 'outer;
             }
@@ -80,7 +97,7 @@ mod peeking {
     fn reads_null_terminated_strings_from_one_word() {
         let data = vec![[102, 111, 111, 0, 0, 0, 0, 0]]
             .into_iter()
-            .map(cast_to_word)
+            .map(cast_to_eight_byte_word)
             .map(Ok);
         assert_eq!(data_to_string(data).unwrap(), b"foo");
     }
@@ -92,7 +109,7 @@ mod peeking {
             [105, 0, 0, 0, 0, 0, 0, 0],
         ]
         .into_iter()
-        .map(cast_to_word)
+        .map(cast_to_eight_byte_word)
         .map(Ok);
         assert_eq!(data_to_string(data).unwrap(), b"abcdefghi");
     }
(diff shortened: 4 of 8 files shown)


src/emulation/mod.rs
```diff
@@ -76,6 +76,24 @@ impl SyscallMock {
                     ptrace::setregs(pid, registers)?;
                 }
             }
+            (Syscall::Stat, SyscallStop::Exit) if !self.protocol.mocked_files.is_empty() => {
+                let filename = tracee_memory::peek_string(pid, registers.rdi)?;
+                if self.protocol.mocked_files.contains(&filename) {
+                    let buffer_ptr = registers.rsi;
+                    #[allow(clippy::forget_copy)]
```
matthandlersux: i'm not sure if this is ok... but clippy is complaining about something that happens after the macro expands :'(

README.md
```diff
@@ -98,6 +98,10 @@ protocols:
         exitcode?: number
           # Mocked exitcode of the command.
           # Default: 0
+        filetree: string
```
soenkehahn: The field should have a `?` at the end.

README.md
```diff
@@ -98,6 +98,10 @@ protocols:
...
+          # List of files and folders that your script expects to exist.
```
soenkehahn: Currently the tool will still find files from the filesystem, right? So maybe this should mention something about mocked files. Maybe `List of files and folders that are going to be mocked to exist.`

README.md
soenkehahn: This should be `[string]`, right?

README.md
```diff
@@ -98,6 +98,10 @@ protocols:
...
+          # Note that directories must include a trailing '/'.
+          # Example: /www/logs/
```
soenkehahn: Maybe `Example: [/www/logs], default: []`?

README.md
soenkehahn: Should we call this `mockedFiles` instead of `filetree`?

src/protocol/mod.rs
```diff
@@ -235,12 +237,23 @@ impl Protocol {
         Ok(())
     }
 
+    fn add_filetree(&mut self, object: &Hash) -> R<()> {
```
soenkehahn: This should be consistent with the field name, i.e. `add_mocked_files`.

src/protocol/mod.rs
```diff
@@ -702,6 +715,24 @@ mod load {
...
+            test_parse_one(
+                r##"
+                    |protocol:
+                    |  - /bin/true
+                    |filetree:
+                    |  - /usr/local/bin/unrar
+                "##
+            )?
+            .mocked_files
+            .map(|mocked_file| String::from_utf8(mocked_file).unwrap()),
```
soenkehahn: Maybe `.map(String::from_utf8_lossy)`?

src/syscall_mock/mod.rs
```diff
@@ -76,6 +76,24 @@ impl SyscallMock {
                     ptrace::setregs(pid, registers)?;
                 }
             }
+            (Syscall::Stat, SyscallStop::Exit) if !self.protocol.mocked_files.is_empty() => {
```
soenkehahn: Woot, woot, pattern guards. Didn't know about those.

soenkehahn: Although, I would remove this guard. As this is just an optimization and adds more complexity.

src/emulation/mod.rs
soenkehahn: Yeah, I think, this is fine. Should probably be fixed in `memoffset`, you could create an issue there.

src/syscall_mock/mod.rs
```diff
@@ -76,6 +76,24 @@ impl SyscallMock {
...
+                let filename = tracee_memory::peek_string(pid, registers.rdi)?;
+                if self.protocol.mocked_files.contains(&filename) {
+                    let buffer_ptr = registers.rsi;
+                    #[allow(clippy::forget_copy)]
+                    let file_mode_offset = offset_of!(libc::stat, st_mode) as u64;
+                    let mock_mode = if filename.ends_with(b"/") {
+                        libc::S_IFDIR as i32
```
soenkehahn: You could move the `as i32` to where `mock_mode` is used. If you want.

src/syscall_mock/mod.rs
```diff
@@ -76,6 +76,24 @@ impl SyscallMock {
                     ptrace::setregs(pid, registers)?;
                 }
             }
+            (Syscall::Stat, SyscallStop::Exit) if !self.protocol.mocked_files.is_empty() => {
+                let filename = tracee_memory::peek_string(pid, registers.rdi)?;
+                if self.protocol.mocked_files.contains(&filename) {
+                    let buffer_ptr = registers.rsi;
```
soenkehahn: I think this variable name should include `stat`. Since that's the name of the struct. Maybe just `statbuf` like in the manpages?

src/syscall_mock/mod.rs
```diff
@@ -76,6 +76,24 @@ impl SyscallMock {
...
+                    #[allow(clippy::forget_copy)]
+                    let file_mode_offset = offset_of!(libc::stat, st_mode) as u64;
```
soenkehahn: I would consider inlining `file_mode_offset`.

src/syscall_mock/mod.rs
```diff
@@ -76,6 +76,24 @@ impl SyscallMock {
...
+                    let mock_mode = if filename.ends_with(b"/") {
+                        libc::S_IFDIR as i32
+                    } else {
+                        libc::S_IFREG as i32
+                    };
+                    tracee_memory::poke_four_bytes(pid, buffer_ptr + file_mode_offset, mock_mode)?;
+
```
soenkehahn: Could you remove this newline, please?

src/tracer/tracee_memory.rs
```diff
@@ -274,6 +311,23 @@ mod poking {
...
+                poke_four_bytes(child, registers.rdi, cast_to_four_byte_word(four_bytes))?;
+
+                assert_eq!(
+                    cast_to_eight_byte_array(peekdata(child, registers.rdi)?),
+                    [40, 30, 20, 10, 5, 6, 7, 8],
+                );
+                Ok(())
+            })
+        }
+
```
soenkehahn: The next test below is a test for the test helper `run_roundtrip_test`. So I think the order of that one and your new test should be switched.

src/tracer/tracee_memory.rs
```diff
@@ -104,13 +121,26 @@ mod peeking {
...
         .into_iter()
-        .map(cast_to_word)
+        .map(cast_to_eight_byte_word)
         .map(Ok);
         assert_eq!(data_to_string(data).unwrap(), b"abcdefgh");
     }
 }
 
-fn pokedata(pid: Pid, address: c_ulonglong, words: c_ulonglong) -> R<()> {
+pub fn poke_four_bytes(pid: Pid, address: c_ulonglong, small_word: i32) -> R<()> {
```
soenkehahn: Maybe for consistency with how the words are passed in change this from `i32` to `c_uint`? (That's an alias for `u32`.)

src/tracer/tracee_memory.rs
```diff
@@ -104,13 +121,26 @@ mod peeking {
...
+    let existing_word: c_ulonglong = peekdata(pid, address)?;
+    let mut overlapping_data_array: [u8; 8] = cast_to_eight_byte_array(existing_word);
+
+    let first_four_bytes: [u8; 4] = cast_to_four_byte_array(small_word);
+    overlapping_data_array[..4].clone_from_slice(&first_four_bytes[..4]);
```
soenkehahn: Is the second `[..4]` necessary? This is only 4 bytes long anyways, right?

src/tracer/tracee_memory.rs
```diff
@@ -104,13 +121,26 @@ mod peeking {
...
+
+    let new_word: c_ulonglong = cast_to_eight_byte_word(overlapping_data_array);
+    ptrace::write(pid, address as *mut c_void, new_word as *mut c_void)?;
```
soenkehahn: Use `pokedata` instead?

src/tracer/tracee_memory.rs
```diff
@@ -274,6 +311,23 @@ mod poking {
             )
         }
 
+        #[test]
+        fn roundtrip_for_four_byte_word_doesnt_clobber_adjacent_four_bytes() -> R<()> {
+            run_roundtrip_test(|child, registers| {
+                let eight_bytes: [u8; 8] = [1, 2, 3, 4, 5, 6, 7, 8];
+                let four_bytes: [u8; 4] = [40, 30, 20, 10];
```
soenkehahn: I think I would use `[10, 20, 30, 40]`.

tests/run.rs
```diff
@@ -800,3 +800,122 @@ fn complains_when_expected_an_unmocked_command() -> R<()> {
...
+                        |error:
+                        |  expected: /bin/cat /foo/bar
+                        |  received: <script terminated>
+                    ",
+                )?),
+            )?;
+            Ok(())
+        }
+    }
+}
```
soenkehahn: @matthandlersux: I would change these integration tests a bit, but I think it'd be easier to talk about it in person, if you're up for that.

src/tracer/tracee_memory.rs
```diff
@@ -3,20 +3,37 @@ use libc::{c_ulonglong, c_void};
...
+fn cast_to_eight_byte_word(bytes: [u8; 8]) -> c_ulonglong {
+    u64::from(bytes[0])
+        + (u64::from(bytes[1]) << 8)
+        + (u64::from(bytes[2]) << 16)
+        + (u64::from(bytes[3]) << 24)
+        + (u64::from(bytes[4]) << 32)
+        + (u64::from(bytes[5]) << 40)
+        + (u64::from(bytes[6]) << 48)
+        + (u64::from(bytes[7]) << 56)
 }
```
soenkehahn: I'm not sure I like this bit-shifting better, but -- as long as we have good test coverage -- I also don't care that much.

tests/run.rs
```diff
@@ -800,3 +800,122 @@ fn complains_when_expected_an_unmocked_command() -> R<()> {
...
+
+    mod in_ruby {
+        use super::*;
+
+        #[test]
+        fn allows_to_mock_files_existence_in_ruby() -> R<()> {
+            test_run(
+                r##"
+                    |#!/usr/bin/env ruby
+                    |`cat /foo/bar` if File.exists?("/foo/bar")
```
soenkehahn: I think I would try to simplify this a bit. These integration tests are not meant to provide reasonable little ruby snippets, but just the most simple test case for what they're testing. So maybe
``` ruby
`/bin/true` if File.exists?("/foo/bar")
```
or even better
``` ruby
fail if File.exists?("./foo")
```

tests/run.rs
```diff
@@ -800,3 +800,122 @@ fn complains_when_expected_an_unmocked_command() -> R<()> {
...
+                    |      - /bin/cat /foo/bar/test
+                    |      - /bin/cat /foo/bar/test2
+                    |    filetree:
+                    |      - /foo/bar/
+                "##,
+                Ok(()),
+            )?;
+            Ok(())
+        }
+    }
```
soenkehahn: If ruby and bash access the file system in the same way, I wouldn't add any ruby tests for this. (I only added one ruby test for spawning subprocesses, since ruby uses different syscalls for that than bash.)

tests/run.rs
```diff
@@ -800,3 +800,122 @@ fn complains_when_expected_an_unmocked_command() -> R<()> {
...
+    mod in_bash {
+        use super::*;
+
+        #[test]
+        fn allows_to_mock_files_existence() -> R<()> {
+            test_run(
+                r##"
+                    |#!/usr/bin/env bash
+                    |if [ -f /foo/bar ]; then
+                    |  cat /foo/bar
```
soenkehahn: Change this to `/bin/true` to make the test simpler?

tests/run.rs
```diff
@@ -800,3 +800,122 @@ fn complains_when_expected_an_unmocked_command() -> R<()> {
...
+            Ok(())
+        }
+
+        #[test]
+        fn allows_to_mock_directory_existence() -> R<()> {
+            test_run(
+                r##"
+                    |#!/usr/bin/env bash
+                    |if [ -d /foo/bar/ ]; then
+                    |  cat /foo/bar/test
```
soenkehahn: Change this to `/bin/true`?

tests/run.rs
```diff
@@ -800,3 +800,122 @@ fn complains_when_expected_an_unmocked_command() -> R<()> {
...
+            Ok(())
+        }
+
+        #[test]
+        fn does_not_mock_existence_of_unspecified_files() -> R<()> {
+            test_run(
+                r##"
+                    |#!/usr/bin/env bash
+                    |if [ -f /foo/bar ]; then
+                    |  cat /foo/bar
```
soenkehahn: Change this to `/bin/true`?

tests/run.rs
```diff
@@ -800,3 +800,122 @@ fn complains_when_expected_an_unmocked_command() -> R<()> {
...
+                    |fi
+                "##,
+                r##"
+                    |protocols:
+                    |  - protocol:
+                    |    - /bin/cat /foo/bar
```
soenkehahn: Change this to `  - protocol: []` and expect the result to be `Ok(())`?

tests/run.rs
soenkehahn: @matthandlersux: Since we didn't get a chance to talk I left some comments. I'll be up for looking at the changed integration tests when you've addressed them.

README.md
matthandlersux: i have no preference, i'll change it

src/protocol/mod.rs
matthandlersux: tried that, was not able to get the types to work out by just passing the function, but 
```rust
.map(|path| String::from_utf8_lossy(&path).to_string()),
```
works. 

if i understand correctly, `&path` takes `Vec<u8>` => `&[u8]` to satisfy `from_utf8_lossy`'s type signature. is there a way around having to explicitly do this? also, `from_utf8_lossy` gives me back a 🐮 `Cow<_, str>` which i think i have to convert in order to compare.

src/syscall_mock/mod.rs
matthandlersux: compromise on `statbuf_ptr`? i think `statbuf_ptr + file_mode_offset` is clearer if the name suggests it's a pointer.

src/tracer/tracee_memory.rs
matthandlersux: i don't feel that strongly, i just thought it would be good to get rid of usage of `unsafe`. let me know if you'd like me to switch back. 

just to be clear, are you asking for more unit tests around these functions?

tests/run.rs
matthandlersux: in other words, it's safe to assume that if ruby uses `stat` and `statbuf` to check for file existence, they would do the same for directories?

tests/run.rs
matthandlersux: note that the tests print out `/tmp/test.1Gf6lsxeepoC/file:2:in `<main>': unhandled exception` if i call `fail` as you suggest. any advice on how to hide that? 

i could always fail if the file *doesn't* exist and instead expect `Ok(())`. i think i like that better because then the polarity of the assertion matches (successful if the feature works vs failure if the feature works), but let me know if you prefer the `fail`.

src/protocol/mod.rs
soenkehahn: Yeah, I think this is all necessary. Instead of `to_string` you could use `into_owned`, I think, which would work for all `Cow`s, not just for `Cow`s of strings.

src/syscall_mock/mod.rs
soenkehahn: Sounds good to me.

src/tracer/tracee_memory.rs
soenkehahn: No, the existing tests are fine. I'm just saying that I don't care that strongly and it's up to you if you want to switch back.

tests/run.rs
soenkehahn: Sorry, I wasn't just talking about the directory tests, but about all ruby tests that you added. So far I only added non-bash tests when other languages behaved differently from bash. In this case it seems that both bash and ruby access the file system in the same way (both for files and directories) so I would consider just sticking to bash in the tests.

tests/run.rs
soenkehahn: No, I don't have a particular preference. `fail` was just an idea. The tests should just be as simple as possible.

tests/run.rs
matthandlersux: got it, that answers my other question then about `fail if...`

src/protocol/mod.rs
```diff
@@ -758,6 +771,24 @@ mod load {
...
+
+    #[test]
+    fn allows_to_specify_mocked_files() -> R<()> {
+        assert_eq!(
+            test_parse_one(
+                r##"
+                    |protocol:
+                    |  - /bin/true
+                    |mockedFiles:
+                    |  - /usr/local/bin/unrar
```
soenkehahn: Maybe use a simpler test protocol:
``` yaml
protocol: []
mockedFiles:
  - /foo
```

PR #128, opened 2019-03-08
implement expected stderr for protocols

### LINKED ISSUE #87: allow to specify the expected stderr of the tested script
It should be possible to capture what the script outputs to `stderr` and compare it to a given expected value.

``` yaml
protocol:
  - foo
  - bar
stderr: expected output
```

soenkehahn:
Fixes #87.

There's UI questions I have about this PR:

1. Currently the error messages assume that the output to `stderr` includes a trailing newline. Otherwise they look weird, like this:
```
fooerror:
  expected output to stderr:
bar  received output to stderr:
foo
```
(The first foo is the actual output of the script that is still being relayed, as intended.) Not too sure what to do about this...

2. Usually error messages should contain a trailing newline. With this PR users *have* to specify the trailing newline explicitly, like this:
``` yaml
protocols:
  - protocol: []
    stderr: "foo\n"
```
That's a bit inconvenient, but I don't really know what to do about this. We don't just want to add a newline, if missing, do we?

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -170,7 +170,7 @@ pub struct Protocol {
     pub arguments: Vec<String>,
     pub env: HashMap<String, String>,
     pub cwd: Option<Vec<u8>>,
-    stderr: Option<Vec<u8>>,
+    pub stderr: Option<Vec<u8>>,
     pub exitcode: i32,
     pub mocked_files: Vec<Vec<u8>>,
 }
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -170,6 +170,7 @@ pub struct Protocol {
     pub arguments: Vec<String>,
     pub env: HashMap<String, String>,
     pub cwd: Option<Vec<u8>>,
+    stderr: Option<Vec<u8>>,
     pub exitcode: i32,
     pub mocked_files: Vec<Vec<u8>>,
 }
@@ -188,6 +189,7 @@ impl Protocol {
             cwd: None,
             exitcode: 0,
             mocked_files: vec![],
+            stderr: None,
         }
     }
 
@@ -230,6 +232,13 @@ impl Protocol {
         Ok(())
     }
 
+    fn add_stderr(&mut self, object: &Hash) -> R<()> {
+        if let Ok(stderr) = object.expect_field("stderr") {
+            self.stderr = Some(stderr.expect_str()?.as_bytes().to_vec());
+        }
+        Ok(())
+    }
+
     fn add_exitcode(&mut self, object: &Hash) -> R<()> {
         if let Ok(exitcode) = object.expect_field("exitcode") {
             self.exitcode = exitcode.expect_integer()?;
@@ -252,6 +261,7 @@ impl Protocol {
         protocol.add_arguments(&object)?;
         protocol.add_env(&object)?;
         protocol.add_cwd(&object)?;
+        protocol.add_stderr(&object)?;
         protocol.add_exitcode(&object)?;
         protocol.add_mocked_files(&object)?;
         Ok(protocol)
@@ -787,6 +797,41 @@ mod load {
             vec![("/foo")]
         );
         Ok(())
+    }
+
+    mod expected_stderr {
+        use super::*;
+        use pretty_assertions::assert_eq;
+
+        #[test]
+        fn allows_to_specify_the_expected_stderr() -> R<()> {
+            assert_eq!(
+                test_parse_one(
+                    r##"
+                        |- protocol: []
+                        |  stderr: foo
+                    "##
+                )?
+                .stderr
+                .map(|s| String::from_utf8(s).unwrap()),
+                Some("foo".to_string())
+            );
+            Ok(())
+        }
+
+        #[test]
+        fn none_is_the_default() -> R<()> {
+            assert_eq!(
+                test_parse_one(
+                    r##"
+                        |- protocol: []
+                    "##
+                )?
+                .stderr,
+                None
+            );
+            Ok(())
+        }
     }
 }
 
--- a/src/syscall_mock/mod.rs
+++ b/src/syscall_mock/mod.rs
@@ -4,6 +4,7 @@ pub mod test_result;
 use crate::context::Context;
 use crate::protocol;
 use crate::protocol::Protocol;
+use crate::tracer::stdio_redirecting::Redirector;
 use crate::tracer::syscall::Syscall;
 use crate::tracer::{tracee_memory, SyscallStop};
 use crate::utils::short_temp_files::ShortTempFile;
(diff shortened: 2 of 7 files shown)


soenkehahn:
@data-pup and I talked about this offline and came up with this:

We'll fix problem 1. by just outputting the captured and expected `stderr` in quotes with escaped newlines. Like this:

```
error:
  expected output to stderr: "foo"
  received output to stderr: "foo\n"
```

And we just won't fix 2. right now. It's a bit inconvenient but at least very straight-forward and self-explanatory. (I.e. if users don't specify a trailing newline they'll get an error message as the one above and then it's pretty obvious that they have to add one.)

This can all be improved upon in future PRs, of course.

matthandlersux:
@soenkehahn cool, that was my first thought too (to put it in quotes and escape the newline). 

as for the first `foo` (that is relayed), you could do what github does and display a hint that indicates "there was no newline at the end of this": 
no newline

but that could come in a later PR where formatted output is improved

src/protocol/mod.rs
```diff
@@ -788,6 +798,41 @@ mod load {
...
+                test_parse_one(
+                    r##"
+                        |- protocol: []
+                    "##
+                )?
+                .stderr,
+                None
+            );
+            Ok(())
+        }
```
data-pup: nice tests :+1:

src/syscall_mock/mod.rs
```diff
@@ -133,26 +136,45 @@ impl SyscallMock {
...
         if exitcode != self.protocol.exitcode {
-            self.register_error(
+            self.register_step_error(
                 &format!("<exitcode {}>", self.protocol.exitcode),
                 &format!("<exitcode {}>", exitcode),
             );
         }
+        if let Some(expected_stderr) = &self.protocol.stderr {
+            match captured_stderr {
+                None => panic!("stderr expected, but not captured"),
```
data-pup: maybe this should register an error similar to the message below? i.e.

```
    expected output to stderr: {:?}
    received no output to stderr
```

src/syscall_mock/mod.rs
```diff
@@ -133,26 +136,45 @@ impl SyscallMock {
...
+                Some(captured_stderr) => {
```
data-pup: not sure if the syntax would allow it, but i believe you could move the `if &captured_stderr != expected_stderr` condition into the pattern match here as a guard.

the benefit would be removing a level of indentation within the block below, but that would also mean adding a `_ => { }` branch at the end so that the arms are comprehensive. Because of that, I'd be happy with either approach here :slightly_smiling_face:

src/syscall_mock/mod.rs
```diff
@@ -133,26 +136,45 @@ impl SyscallMock {
...
+                    if &captured_stderr != expected_stderr {
+                        self.register_error(format!(
+                            "  expected output to stderr: {:?}\
+                             \n  received output to stderr: {:?}\n",
+                            String::from_utf8_lossy(&expected_stderr.clone()).as_ref(),
+                            String::from_utf8_lossy(&captured_stderr.clone()).as_ref(),
```
data-pup: is `&captured_stderr.clone()` redundant? or is there some type system stuff (`Cow` perhaps?) preventing that?

src/syscall_mock/mod.rs
```diff
@@ -133,26 +136,45 @@ impl SyscallMock {
...
+                        ));
+                    }
+                }
+            }
+        }
         self.result
     }
 
-    fn register_error(&mut self, expected: &str, received: &str) {
+    fn register_step_error(&mut self, expected: &str, received: &str) {
```
data-pup: I like the renaming here :slightly_smiling_face:

Style question: I agree that using borrowed values like `&str` is ideal when choosing a method's signature, but it seems like we are almost(?) always passing a `&format(..)` in as the parameter. 

Maybe this function could accept `String` values for `expected` and `received`? The benefit would be less syntax noise at the call sites, but that would mean this would be the terminal point in those values' lifetimes. Happy with either approach here, I'd say this is also a stylistic option :slightly_smiling_face:

tests/stdio.rs
```diff
@@ -54,3 +54,47 @@ fn relays_stderr_from_the_tested_script_to_the_user() -> R<()> {
...
+            "##,
+            r##"
+                |protocols:
+                |  - protocol: []
+                |    stderr: "foo\n"
+            "##,
+            Ok(()),
+        )?;
+        Ok(())
+    }
```
data-pup: Maybe add a test `fails_when_nothing_captured`?

src/tracer/stdio_redirecting.rs
```diff
@@ -2,20 +2,30 @@ use crate::context::Context;
...
 
 type RawFd = c_int;
 
 pub struct Redirector {
     stdout: Redirect,
-    stderr: Redirect,
+    pub stderr: Redirect,
+}
+
+pub enum CaptureStderr {
```
data-pup: I really like the idea behind this `enum` :+1:

src/tracer/stdio_redirecting.rs
```diff
@@ -42,21 +52,39 @@ enum StreamType {
     Stderr,
 }
 
-struct Redirect {
+pub struct Redirect {
     stream_type: StreamType,
     context: Context,
     read_end: RawFd,
     write_end: RawFd,
+    captured: Option<Arc<Mutex<Cursor<Vec<u8>>>>>,
```
data-pup: Nice! `Now<We<Are<Writing<Concurrent<Code<_>>>>>>`

src/syscall_mock/mod.rs
soenkehahn: This only happens if `check-protocols` has a bug: when the expected `stderr` is given in the protocol, but the script is started without capturing it. That's also why it's a `panic` instead of an `Err`. How can we make this more obvious? Maybe `panic!("check-protocols bug: stderr expected, but not captured")`?

data-pup: :open_mouth: :bulb:  aaaaaaah, that makes sense! I like that error message you proposed, that helps clarify that it's an internal problem rather than something related to the behavior of the subject under test :)

src/syscall_mock/mod.rs
soenkehahn: Syntactically that'd be allowed. I tried it out and it felt less clear than the current version though.

data-pup: makes sense to me! can already imagine how terse that would look. I'll mark this as resolved :+1:

src/syscall_mock/mod.rs
soenkehahn: Yes, good call.

src/syscall_mock/mod.rs
soenkehahn: I tried that, it results in less `&`s but two more `.to_string()`s. Given that `&str` as an argument feels more natural, I'll go with that.

tests/stdio.rs
soenkehahn: Done.

PR #139, opened 2019-03-11
allow to create protocols by running the tested scripts

### LINKED ISSUE #85: allow to create protocols by running the tested scripts
It should be possible to create a protocols file by running `check-protocols` against a script file somehow. It would capture the issued commands and put them into the protocol.

soenkehahn:
Fixes #85.


src/tracer/mod.rs
```diff
@@ -111,17 +125,17 @@ impl Tracer {
...
         args: Vec<String>,
         env: HashMap<String, String>,
         capture_stderr: CaptureStderr,
-        mk_syscall_mock: F,
-    ) -> R<TestResult>
+        mk_syscall_mock: MkSyscallMock,
+    ) -> R<MockResult>
     where
-        F: FnOnce(Pid) -> SyscallMock,
+        MkSyscallMock: FnOnce(Pid) -> Box<SyscallMock<Result = MockResult>>, // fixme
```
data-pup: // fixme :grimacing:

src/recorder.rs
```diff
@@ -0,0 +1,41 @@
...
+use libc::user_regs_struct;
+use nix::unistd::Pid;
+use yaml_rust::Yaml;
+
+pub struct Recorder {
+    protocol: Protocol,
+}
+
+impl Recorder {
+    pub fn new() -> Recorder {
```
data-pup: maybe impl `Default` for this instead of a `new` constructor?

src/protocol/mod.rs
```diff
@@ -172,7 +172,7 @@ pub struct Protocol {
     pub env: HashMap<String, String>,
     pub cwd: Option<Vec<u8>>,
     pub stderr: Option<Vec<u8>>,
-    pub exitcode: i32,
+    pub exitcode: Option<i32>,
```
data-pup: A+

src/cli.rs
```diff
@@ -1,22 +1,120 @@
...
+            ))
+            .arg(
+                Arg::with_name("program")
+                    .help("the program to test")
+                    .required(true)
+                    .index(1),
+            )
+            .get_matches_from_safe(args)?;
+        Ok(Args::CheckProtocols {
+            script_path: PathBuf::from(matches.value_of("program").unwrap()),
```
matthandlersux: it's a force of habit of mine from scala, and i'm not sure if it fully carries over to rust, but even though the arg is required and it should be safe, i would still consider using `ok_or` over `unwrap`.

src/lib.rs
```diff
@@ -71,12 +75,21 @@ mod wrap_main {
...
 
-pub fn run_main(context: &Context, args: impl Iterator<Item = String>) -> R<ExitCode> {
-    Ok(match cli::parse_args(args)? {
+pub fn run_main(context: &Context, args: &cli::Args) -> R<ExitCode> {
+    Ok(match args {
         cli::Args::ExecutableMock {
             executable_mock_path,
         } => executable_mock::run(context, &executable_mock_path)?,
-        cli::Args::CheckProtocols { script_path } => run_check_protocols(context, &script_path)?,
+        cli::Args::CheckProtocols {
```
matthandlersux: not sure if it's an improvement, but reminder that you can do
```rust
        cli::Args::CheckProtocols {
            script_path,
            record: true,
        } => print_recorded_protocol(context, script_path)?,
        cli::Args::CheckProtocols { script_path, .. } => {
            run_check_protocols(context, &script_path)?
        }
```

soenkehahn: Good idea. I tried that and I actually think it looks slightly worse, though.

PR #144, opened 2019-03-14
record command exitcodes

soenkehahn:


soenkehahn:
I still want to add a test for piping a command into an unmocked command and make sure that the concurrency doesn't confuse the `Recorder`.

soenkehahn:
Supporting concurrency is a bigger endeavour. That currently doesn't even work for checking existing protocols. So I'll merge this PR without that.

PR #154, opened 2019-03-15
clear terminal and terminal scrollback during 'just dev'

soenkehahn:
@hallettj Can you make sure this works on your setup, please?


hallettj:
@soenkehahn Yes, it works nicely.

PR #159, opened 2019-03-18
support process environments when recording protocols 

### LINKED ISSUE #147: allow to specify an environment for the tested script when recording protocols
There's two options for the UI for this:

1. Have a cli flag, e.g. `--env FOO=BAR --env BAZ=HUHU`

2. Have some mechanism to put holes into protocols, e.g.:

``` yaml
env:
  FOO: BAR
  BAZ: HUHU
protocols:
  - protocol:
      - _
```

Here, the underscore is used to mean that `check-protocols` should fill in the rest.

soenkehahn:
Fixes #147.

### DIFF
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -305,6 +305,16 @@ impl Protocol {
         Ok(protocol)
     }
 
+    fn serialize_env(&self, object: &mut Hash) {
+        if !self.env.is_empty() {
+            let mut env = LinkedHashMap::new();
+            for (key, value) in &self.env {
+                env.insert(Yaml::from_str(key), Yaml::from_str(value));
+            }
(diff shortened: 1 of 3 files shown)


src/protocol/mod.rs
```diff
@@ -1029,6 +1040,13 @@ mod serialize {
         }]);
         roundtrip(Protocols::new(vec![protocol]))
     }
+
+    #[test]
+    fn includes_the_environment() -> R<()> {
+        let mut protocol = Protocol::empty();
+        protocol.env.insert("FOO".to_string(), "BAR".to_string());
```
astampoulis: `FOO` is the environment variable and `BAR` is the value? If so, I'd do `bar` instead for the value.

soenkehahn: Done.

PR #162, opened 2019-03-19
make debugging output for protocol structs better readable

### LINKED ISSUE #149: better `Debug` output for protocol types
Currently the strings that the `Debug` trait generates for the protocol types is fairly unreadable, e.g.:
`Command { executable: [47, 117, 115, 114, 47, 98, 105, 110, 47, 106, 113], arguments: [[46]] }`

This should be fixed.

soenkehahn:
Fixes #149.

### DIFF
--- a/src/protocol/command.rs
+++ b/src/protocol/command.rs
@@ -2,7 +2,6 @@ use super::argument_parser::Parser;
 use super::executable_path;
 use crate::R;
 use std::ffi::OsString;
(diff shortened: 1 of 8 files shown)


soenkehahn:
I didn't change `Step.stdout` and `Protocol.stderr`, because those strings could theoretically contain null bytes and those are not allowed in `OsString` or `PathBuf`.

PR #166, opened 2019-03-20
convert "issue board" link into a badge as well

matthandlersux:
might as well go full badge


matthandlersux:
also, fyi, i just uninstalled the wafflebot from this repo (after posting and seeing it was still connected 😢 )

soenkehahn:
Thanks!

PR #168, opened 2019-03-20
use &mut Write instead of Box<Write> in write_yaml

soenkehahn:
I think this is more idiomatic.

### DIFF
--- a/src/protocol/yaml.rs
+++ b/src/protocol/yaml.rs
@@ -110,7 +110,7 @@ fn adjust_yaml_output(input: Vec<u8>) -> Vec<u8> {
     result
 }
 
-pub fn write_yaml(mut output_stream: Box<io::Write>, yaml: &Yaml) -> R<()> {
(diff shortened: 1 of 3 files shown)


src/protocol/yaml.rs
```diff
@@ -92,7 +92,7 @@ fn adjust_yaml_output(input: Vec<u8>) -> Vec<u8> {
     result
 }
 
-pub fn write_yaml(mut output_stream: Box<io::Write>, yaml: &Yaml) -> R<()> {
+pub fn write_yaml(output_stream: &mut io::Write, yaml: &Yaml) -> R<()> {
```
data-pup: https://doc.rust-lang.org/edition-guide/rust-2018/trait-system/dyn-trait-for-trait-objects.html

```
Box<Trait> becomes Box<dyn Trait>
&Trait and &mut Trait become &dyn Trait and &mut dyn Trait
```

This looks great! If possible, I might recommend `&mut dyn io::Write` instead :question:

soenkehahn: Done.

PR #170, opened 2019-03-21
ability to test commands with regex matching

### LINKED ISSUE #122: ability to test commands with generated/variable arguments
i have a script that writes to a logfile and it includes the date, so right now i'm not able to write a protocol that passes. here is an example of the error i get: 
```
error:
  expected: /bin/sh -c "echo \"2019-03-06T22:15:12+00:00 - start: /directory - filename - test\" >> /Users/testuser/tmp/unrar.log"
  received: /bin/sh -c "echo \"2019-03-06T22:16:05+00:00 - start: /directory - filename - test\" >> /Users/testuser/tmp/unrar.log"
```
i imagine the ability to place a regex in my command is a straightforward way to solve this, any other ideas?

matthandlersux:
this PR updates `Step` so that it can take either a `Command` (as it previously did) or a `Regex`, and then it can more loosely match commands (for example, one command might include the current time).

closes #122

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,15 @@ protocols:
       # Default: 0.
     protocol:
       # List of commands that your script is expected to execute.
-      - command: string
-          # the executable, followed by its arguments, separated by spaces.
+      - command|regex: string
+          # One of either `command` or `regex` is required
+          #
+          # command: the executable, followed by its arguments, separated by spaces.
           # Example: /bin/chmod +x foo.sh
+          #
+          # regex: a regular expression (for valid syntax, see: https://docs.rs/regex/1.1.2/regex/#syntax)
+          # Note that the regex is automatically anchored, so it must match the entire command and its arguments
+          # Example: /bin/echo \d+
         stdout?: string
           # Mocked output of this command.
           # Default: ""
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
-# Check Protocols
-[![CircleCI](https://circleci.com/gh/Originate/check-protocols.svg?style=svg)](https://circleci.com/gh/Originate/check-protocols)
-[![open issues](https://img.shields.io/github/issues/originate/check-protocols.svg?label=open%20issues)](https://github.com/Originate/check-protocols/issues)
-[![open pull requests](https://img.shields.io/github/issues-pr-raw/originate/check-protocols.svg?color=%238888ff)](https://github.com/Originate/check-protocols/pulls)
-[![closed issues](https://img.shields.io/github/issues-closed/originate/check-protocols.svg?color=green&label=closed%20issues)](https://github.com/Originate/check-protocols/issues?q=is%3Aissue+is%3Aclosed)
-[![issue board](https://img.shields.io/badge/check--protocols-issue%20board-important.svg)](https://github.com/orgs/Originate/projects/1)
+# `scriptkeeper`
+[![CircleCI](https://circleci.com/gh/Originate/scriptkeeper.svg?style=svg)](https://circleci.com/gh/Originate/scriptkeeper)
+[![open issues](https://img.shields.io/github/issues/originate/scriptkeeper.svg?label=open%20issues)](https://github.com/Originate/scriptkeeper/issues)
+[![open pull requests](https://img.shields.io/github/issues-pr-raw/originate/scriptkeeper.svg?color=%238888ff)](https://github.com/Originate/scriptkeeper/pulls)
+[![closed issues](https://img.shields.io/github/issues-closed/originate/scriptkeeper.svg?color=green&label=closed%20issues)](https://github.com/Originate/scriptkeeper/issues?q=is%3Aissue+is%3Aclosed)
+[![issue board](https://img.shields.io/badge/scriptkeeper-issue%20board-important.svg)](https://github.com/orgs/Originate/projects/1)
 
 Run tests against your scripts without changing your scripts.
 
@@ -13,9 +13,9 @@ Automated tests help us write well-behaved applications, and that's great, but
 what about all those pesky little scripts we use in and around our applications
 (e.g. deploy scripts)? How do we test those?
 
-`check-protocols` is a tool for people who wish to write tests for existing
+`scriptkeeper` is a tool for people who wish to write tests for existing
 scripts and/or use TDD to write new scripts. Because of its design,
-`check-protocols` is language agnostic, since it mocks out syscalls. That means
+`scriptkeeper` is language agnostic, since it mocks out syscalls. That means
 you can test Bash scripts just as well as Python, Ruby, etc.
 
 
@@ -24,14 +24,14 @@ you can test Bash scripts just as well as Python, Ruby, etc.
 **This tool is very experimental. It may give incorrect results and delete your
 files.**
 
-There are lots and lots of features still missing from `check-protocols`. If you
+There are lots and lots of features still missing from `scriptkeeper`. If you
 try it out, I'd be interested to hear which features you would want the most.
 Feel free to open (or vote on)
-[issues](https://github.com/Originate/check-protocols/issues).
+[issues](https://github.com/Originate/scriptkeeper/issues).
 
 ## Usage
 
-`check-protocols` allows you to write tests -- so-called protocols -- for
+`scriptkeeper` allows you to write tests -- so-called protocols -- for
 scripts (or other executables) -- without the need to modify your executables.
 
 Here's an example script `./build-image.sh`:
@@ -65,7 +65,7 @@ protocols:
     exitcode: 1
 ```
 
-Now running `check-protocols ./build-image.sh` will tell you whether your script
+Now running `scriptkeeper ./build-image.sh` will tell you whether your script
 `./build-image.sh` conforms to your protocols in
 `./build-image.sh.protocols.yaml`.
 
@@ -88,7 +88,7 @@ protocols:
       { [string]: string }
     cwd?: string
       # Current working directory the tested script will be executed in.
-      # Example: /test-dir, default: same directory that `check-protocols` is run in.
+      # Example: /test-dir, default: same directory that `scriptkeeper` is run in.
     mockedFiles?: [string]
       # List of files and folders that are going to be mocked to exist.
       # Note that directories must include a trailing '/'.
@@ -192,8 +192,8 @@ This will actually execute the sub-commands that your script performs, without
 mocking them out. And it will overwrite your protocols file with the recorded
 version.
 
-You can also start with a partial protocol and have `check-protocols` fill in
-the specified holes:
+You can also start with a partial protocol and have `scriptkeeper` fill in the
+specified holes:
 
 ``` yaml
 protocols:
@@ -206,7 +206,7 @@ protocols:
 This allows for an iterative process to create a protocol:
 
 1. Start with an empty protocol with a hole.
-2. Run `check-protocols`.
+2. Run `scriptkeeper`.
 3. Identify the step in the recorded protocol where it deviates from the
    intended test. (If it doesn't, you're done.)
 4. Refine the protocol by modifying the inputs to the tested script, i.e. the
@@ -222,7 +222,7 @@ You can run the tool inside docker, for example like this:
 
 ``` bash
 ./build-docker-image.sh
-./check-protocols-in-docker.sh <PATH_TO_YOUR_SCRIPT>
+./scriptkeeper-in-docker.sh <PATH_TO_YOUR_SCRIPT>
 ```
 
 ## Contributing
--- a/src/protocol/command.rs
+++ b/src/protocol/command.rs
@@ -20,6 +20,11 @@ impl Command {
         }
     }
 
+    pub fn compare(&self, other: &Command) -> bool {
+        executable_path::compare_executables(&self.executable, &other.executable)
+            && self.arguments == other.arguments
+    }
+
     fn escape(word: String) -> String {
         fn escape_char(char: char) -> String {
             match char {
--- a/src/protocol/command.rs
+++ b/src/protocol/command.rs
@@ -6,6 +6,32 @@ use std::path::PathBuf;
 use std::str;
 
 #[derive(PartialEq, Eq, Debug, Clone)]
+pub enum CommandMatcher {
+    Exact(Command),
+}
+
+impl CommandMatcher {
+    pub fn matches_received(&self, received: &Command) -> bool {
+        match self {
+            CommandMatcher::Exact(self_command) => {
+                executable_path::compare_executables(&self_command.executable, &received.executable)
+                    && self_command.arguments == received.arguments
+            }
+        }
+    }
+
+    pub fn format(&self) -> String {
+        match self {
+            CommandMatcher::Exact(self_command) => self_command.format(),
+        }
+    }
+
+    pub fn exact_match(command: &str) -> R<CommandMatcher> {
+        Ok(CommandMatcher::Exact(Command::new(command)?))
+    }
+}
+
+#[derive(PartialEq, Eq, Debug, Clone)]
 pub struct Command {
     pub executable: PathBuf,
     pub arguments: Vec<OsString>,
@@ -18,11 +44,6 @@ impl Command {
         } else {
             word
         }
-    }
-
-    pub fn compare(&self, other: &Command) -> bool {
-        executable_path::compare_executables(&self.executable, &other.executable)
-            && self.arguments == other.arguments
     }
 
     fn escape(word: String) -> String {
--- a/src/protocol/command.rs
+++ b/src/protocol/command.rs
@@ -1,33 +1,92 @@
 use super::argument_parser::Parser;
 use super::executable_path;
 use crate::R;
+use regex::Regex;
 use std::ffi::OsString;
 use std::path::PathBuf;
 use std::str;
 
-#[derive(PartialEq, Eq, Debug, Clone)]
+#[derive(Debug, Clone)]
 pub enum CommandMatcher {
     Exact(Command),
+    RegexMatch(Regex),
 }
+
+impl PartialEq for CommandMatcher {
+    fn eq(&self, other: &CommandMatcher) -> bool {
+        use CommandMatcher::*;
+        match (self, other) {
+            (Exact(a), Exact(b)) => a == b,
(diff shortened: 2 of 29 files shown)


README.md
```diff
@@ -101,9 +101,15 @@ protocols:
       # Default: 0.
     protocol:
       # List of commands that your script is expected to execute.
-      - command: string
-          # the executable, followed by its arguments, separated by spaces.
+      - command|regex: string
```
matthandlersux: i wasn't sure how to format the idea of "either" since we have a sort of pseudo syntax here, thoughts?

src/protocol/command.rs
```diff
@@ -1,10 +1,95 @@
...
+    pub fn format(&self) -> String {
+        match self {
+            CommandMatcher::Exact(command) => command.format(),
+            CommandMatcher::RegexMatch(regex) => regex.as_str().to_string(),
+        }
+    }
+
+    pub fn regex_match(string: &str) -> R<CommandMatcher> {
+        Ok(CommandMatcher::RegexMatch(Regex::new(&format!(
+            "^{}$",
```
matthandlersux: unfortunately the `Regex` create doesn't offer control over anchoring via the `RegexBuilder`, so i couldn't find any other way to get this in there. FWIW, even if the protocol includes anchors, the `Regex` ignores consecutive ones (ie, `^^\d$$` seems to be the same as `^\d$`)

src/protocol/mod.rs
```diff
@@ -20,13 +20,13 @@ use yaml_rust::{yaml::Hash, Yaml, YamlLoader};
 
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub struct Step {
-    pub command: Command,
+    pub command: CommandMatcher,
```
matthandlersux: i held off on renaming this field because i want to see if we prefer something other than `CommandMatcher`. once we settle on that though, i imagine this field name should update as well.

src/protocol/mod.rs
```diff
@@ -1023,6 +1037,51 @@ mod load {
...
+                r"
+                    |protocols:
+                    |  - protocol:
+                    |      - regex: \d
+                ",
+            )?
+            .steps[0]
+                .clone();
+            match step.command {
+                CommandMatcher::RegexMatch(regex) => assert_eq!(regex.as_str(), "^\\d$"),
```
matthandlersux: i'm not in love with the fact that i had to inject the anchors in here, but i was trying to avoid using `CommandMatcher::regex_match` to build this based on our conversation today.

src/protocol/command.rs
```diff
@@ -1,10 +1,95 @@
...
+            (RegexMatch(a), RegexMatch(b)) => a.as_str() == b.as_str(),
+            _ => false,
+        }
+    }
+}
+
+impl Eq for CommandMatcher {}
+
+impl CommandMatcher {
+    pub fn matches_received(&self, received: &Command) -> bool {
```
soenkehahn: I think it's confusing that this has `received` in its name. I think from the perspective of the `CommandMatcher` it doesn't matter that this is used to match commands that are received. I think this should just be `matches` or `matches_command`.

src/protocol/command.rs
soenkehahn: Can we have a test (or two) for the claim that it ignores consecutive anchors?

src/protocol/command.rs
```diff
@@ -1,10 +1,95 @@
...
+    pub fn regex_match(string: &str) -> R<CommandMatcher> {
+        Ok(CommandMatcher::RegexMatch(Regex::new(&format!(
+            "^{}$",
+            string
+        ))?))
+    }
+
+    pub fn exact_match(command: &str) -> R<CommandMatcher> {
+        Ok(CommandMatcher::Exact(Command::new(command)?))
+    }
```
soenkehahn: I was a bit confused that `regex_match` and `exact_match` are constructors.

- I think they should be at the top of this impl block.
- Maybe change their names to (`new_regex` or `from_regex`) and `new_exact`?

src/protocol/command.rs
```diff
@@ -1,10 +1,95 @@
...
+        Ok(())
+    }
+
+    #[test]
+    fn regex_only_matches_if_entire_command_is_a_match() -> R<()> {
+        assert!(!CommandMatcher::regex_match("cp \\d")?.matches_received(&Command::new("cp 1 2")?));
+        Ok(())
+    }
+}
+
```
soenkehahn: @matthandlersux: Since you mentioned that you like small files: I'd be fine if this were in a separate module.

src/protocol/command.rs
```diff
@@ -1,10 +1,95 @@
...
 use crate::R;
+use regex::Regex;
 use std::ffi::OsString;
 use std::path::PathBuf;
 use std::str;
 
+#[derive(Debug, Clone)]
+pub enum CommandMatcher {
+    Exact(Command),
+    RegexMatch(Regex),
```
soenkehahn: I wonder if it'd be better for multiple reasons to have `RegexMatch(String)` instead.

- Then we could definitely derive `PartialEq` and `Eq`.
- Also, we could then consider getting rid of the constructor functions below and just use `Exact(Command::new(...))` and `RegexMatch(...)`. And make those public as needed.

I think making enums part of an interface us often a good idea, because it gives developers a quick way to gain a lot of understanding about how something works.

src/protocol/mod.rs
soenkehahn: Yeah, I wouldn't use `regex_match`. If we make the change to `RegexMatch(String)` then we would leave out the anchors of the `String`. That way our in-memory representation is very close to the given yaml, which I kind of like.

src/protocol/mod.rs
```diff
@@ -1023,6 +1037,51 @@ mod load {
...
+        fn parses_a_regex_command_matcher() -> R<()> {
+            let step = test_parse_one(
+                r"
+                    |protocols:
+                    |  - protocol:
+                    |      - regex: \d
+                ",
+            )?
+            .steps[0]
+                .clone();
```
soenkehahn: I think you could do `.get(0).unwrap()`.

src/protocol/mod.rs
```diff
@@ -1023,6 +1037,51 @@ mod load {
...
+        fn disallows_a_regex_matcher_and_command() -> R<()> {
+            let tempfile = TempFile::new()?;
+            assert_error!(
+                test_parse(
+                    &tempfile,
+                    r"
+                        |protocols:
+                        |  - protocol:
+                        |      - regex: \d
+                        |        command: foo
```
soenkehahn: `command` before `regex`?

src/protocol/mod.rs
```diff
@@ -1023,6 +1037,51 @@ mod load {
...
+                ),
+                format!(
+                    "error in {}.protocols.yaml: \
+                     please provide either a 'command' or 'regex' but not both",
+                    path_to_string(&tempfile.path())?
+                )
+            );
+            Ok(())
+        }
+    }
```
soenkehahn: Very nice tests.

tests/run.rs
```diff
@@ -104,6 +104,23 @@ fn can_specify_interpreter() -> R<()> {
     Ok(())
 }
 
+#[test]
+fn can_specify_regex() -> R<()> {
```
soenkehahn: Maybe `allows_to_match_commands_by_regex`?

tests/run.rs
```diff
@@ -104,6 +104,23 @@ fn can_specify_interpreter() -> R<()> {
...
+        ",
+        r#"
+            |protocols:
+            |  - protocol:
+            |    - regex: cp foo \w+
+        "#,
+        Ok(()),
+    )?;
+    Ok(())
+}
```
soenkehahn: Should we add one more test that shows that regexes can be used to match on the command, not just the arguments?

src/protocol/mod.rs
soenkehahn: I like `CommandMatcher`.

src/protocol/command.rs
```diff
@@ -1,10 +1,95 @@
...
+        use CommandMatcher::*;
+        match (self, other) {
+            (Exact(a), Exact(b)) => a == b,
+            (RegexMatch(a), RegexMatch(b)) => a.as_str() == b.as_str(),
+            _ => false,
+        }
+    }
+}
+
+impl Eq for CommandMatcher {}
```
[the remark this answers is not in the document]
matthandlersux: `Regex` doesn't have an `Eq` or `PartialEq` impl

src/protocol/mod.rs
matthandlersux: i had that but clippy yelled at me

src/protocol/command.rs
matthandlersux: this makes sense to me, but there is a snag, maybe you have some advice for: 

`Regex::new` returns a `Result` which will error if there is an issue with the regex. IMO this error should occur at YAML parse time (i'll add a test for this). the options i can think of are:
* defer regex errors to run time (so the protocol test will always fail)
* run `Regex::new` at construction, bubble up the error, ignore the `Ok` result, reconstruct it at match time
* run `Regex::new` at construction, bubble up the error, save the `Ok` result
  * in the `enum` like `RegexMatch(String, Regex)` (but this doesn't fix the `PartialEq` issue)
  * in the `enum` like `RegexMatch(String, Box<Regex>)` (guessing here, not sure if this fixes the `PartialEq` issue)
  * using `lazy_static` as recommended here: https://github.com/rust-lang/regex#usage-avoid-compiling-the-same-regex-in-a-loop , but i have to doublecheck on how that actually works, and if it would work here

src/protocol/mod.rs
matthandlersux: should i change the name of this field from `command` => `command_matcher`? just want to doublecheck. if yes, would you like me to do anything extra WRT backwards compatibility?

soenkehahn: Yes, please change.

You're asking about backwards compatibility of the library? I don't care about that at all.

src/protocol/command.rs
soenkehahn: I see. :| I think the error should happen at parse time, so I don't think we should defer that.

The second idea (calling `Regex::new` twice) is a bit wonky.

Using `Box<Regex>` doesn't solve the issue, see [here](https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-PartialEq%3CBox%3CT%3E%3E).

Using `lazy_static` feels even wonkier than calling it twice, I think.

What about this: Create a thin wrapper around `Regex` that has the original string and the compiled `Regex`. With `YourCoolWrapper::new() -> R<YourCoolWrapper>`. And then for that we implement `PartialEq` and `Eq` manually. But then for `CommandMatcher` we can derive it. Maybe it can be called `AnchoredRegex`, since that's the functionality that it'd add.

matthandlersux: i like that, will do

matthandlersux:
@soenkehahn i believe i have addressed all comments

src/protocol/mod.rs
```diff
@@ -56,8 +58,15 @@ impl Step {
...
-                check_keys(&["command", "stdout", "exitcode"], object)?;
-                let mut step = Step::from_string(object.expect_field("command")?.expect_str()?)?;
+                check_keys(&["command", "stdout", "exitcode", "regex"], object)?;
+                let mut step = match (object.expect_field("command"), object.expect_field("regex"))
+                {
+                    (Ok(command_field), Err(_)) => Step::from_string(command_field.expect_str()?)?,
+                    (Err(_), Ok(regex_field)) => Step::new(CommandMatcher::RegexMatch(
+                        AnchoredRegex::new(regex_field.expect_str()?)?,
+                    )),
+                    _ => Err("please provide either a 'command' or 'regex' but not both")?,
```
soenkehahn: Maybe `please provide either a 'command' or 'regex' field but not both`?

src/protocol/mod.rs
```diff
@@ -609,6 +618,13 @@ mod load {
         fn multiple_unknown_fields() {}
     }
 
+    fn get_exact(step: Step) -> Command {
+        match step.command_matcher {
+            CommandMatcher::ExactMatch(command) => command,
+            CommandMatcher::RegexMatch(_) => panic!("expected Exact"),
+        }
+    }
```
soenkehahn: Nice.

tests/run.rs
```diff
@@ -104,6 +104,40 @@ fn can_specify_interpreter() -> R<()> {
...
+        ",
+        r#"
+            |protocols:
+            |  - protocol:
+            |    - regex: \w+ foo bar
+        "#,
+        Ok(()),
+    )?;
+    Ok(())
+}
```
soenkehahn: Do you want to put these two tests into a submodule?

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub fn is_match(&self, other: &str) -> bool {
+        self.1.is_match(other)
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum CommandMatcher {
+    ExactMatch(Command),
+    RegexMatch(AnchoredRegex),
+}
```
soenkehahn: Maybe move this type to the top of the file.

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub fn format(&self) -> String {
+        match self {
+            CommandMatcher::ExactMatch(command) => command.format(),
+            CommandMatcher::RegexMatch(AnchoredRegex(string, _)) => string.clone(),
+        }
+    }
+}
+
+#[cfg(test)]
+mod command_matcher {
```
soenkehahn: In all these tests use the simplest examples that test what the description says it tests.

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn exact_command_matches_received() -> R<()> {
+        assert!(CommandMatcher::ExactMatch(Command::new("cp ./")?).matches(&Command::new("cp ./")?));
+        Ok(())
+    }
+
+    #[test]
+    fn exact_command_doesnt_match_different_command() -> R<()> {
+        assert!(!CommandMatcher::ExactMatch(Command::new("cp ./")?).matches(&Command::new("bar")?));
+        Ok(())
+    }
```
soenkehahn: Maybe it'd be nicer to have comprehensive tests for `CommandMatcher` here and get rid of some of the integration tests that test this functionality now.

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn regex_still_matches_if_front_anchor_is_included() -> R<()> {
+        assert!(test_regex_matches_command("^cp \\d", "cp 1")?);
+        Ok(())
+    }
+
+    #[test]
+    fn regex_still_matches_if_end_anchor_is_included() -> R<()> {
+        assert!(test_regex_matches_command("cp \\d$", "cp 1")?);
+        Ok(())
+    }
```
soenkehahn: Maybe put the exact and the regex tests in submodules?

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,153 @@
...
+            Ok(())
+        }
+
+        #[test]
+        fn matches_command_even_if_it_doesnt_exist() -> R<()> {
+            assert!(CommandMatcher::ExactMatch(Command::new("foo")?).matches(&Command::new("foo")?));
+            Ok(())
+        }
+
+        #[test]
```
matthandlersux: @soenkehahn i tried to add the extra tests we talked about, but let me know if i missed one

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,153 @@
...
+        fn doesnt_match_a_different_command() -> R<()> {
+            assert!(
+                !CommandMatcher::ExactMatch(Command::new("foo")?).matches(&Command::new("bar")?)
+            );
+            Ok(())
+        }
+
```
soenkehahn: Please, remove this newline.

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,153 @@
...
+        #[test]
+        fn matches_command_with_full_path() -> R<()> {
+            assert!(CommandMatcher::ExactMatch(Command::new("/bin/true")?)
+                .matches(&Command::new("/bin/true")?));
+            Ok(())
+        }
+
+        #[test]
+        fn doesnt_match_a_different_command() -> R<()> {
```
soenkehahn: Maybe add a test for having the same executable but differing arguments?

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,153 @@
...
+                Ok(())
+            }
+
+            #[test]
+            fn matches_if_end_anchor_is_included() -> R<()> {
+                assert!(test_regex_matches_command("cp$", "cp")?);
+                Ok(())
+            }
+        }
+
```
soenkehahn: Remove newline.

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,153 @@
...
+        #[test]
+        fn doesnt_match_a_command_if_regex_doesnt_match() -> R<()> {
+            assert!(!test_regex_matches_command("foo", "bar")?);
+            Ok(())
+        }
+
+        #[test]
+        fn only_matches_if_entire_command_is_a_match() -> R<()> {
+            assert!(!test_regex_matches_command("cp \\d", "cp 1 2")?);
```
soenkehahn: Maybe `assert!(!test_regex_matches_command("cp", "cp foo")?);`?

src/protocol/command_matcher.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        fn test_regex_matches_command(regex: &str, command: &str) -> R<bool> {
+            let result = CommandMatcher::RegexMatch(AnchoredRegex::new(regex)?)
+                .matches(&Command::new(command)?);
+            Ok(result)
+        }
+
+        #[test]
+        fn matches_a_command() -> R<()> {
+            assert!(test_regex_matches_command("cp", "cp")?);
```
soenkehahn: I think this test should demonstrate actually using a regex. So something like `assert!(test_regex_matches_command("cp foo_bar", "cp .*_bar")?);`?

matthandlersux:
getinsertpic.com

PR #174, opened 2019-03-22
narrow tested properties in cli tests

soenkehahn:
What do people think of this? This is using a tiny crate that I came up with yesterday: [unsafe_get](https://github.com/soenkehahn/unsafe_get). We could also just put this into this project. But it seems generally useful. Should I put it on crates.io?

@matthandlersux @jondistad @data-pup @hallettj


hallettj:
I think it could be useful to have that crate available.

data-pup:
this is really neat, I would endorse that as a separate crate for sure :slightly_smiling_face:

Issue #114, opened 2019-03-04
Allow mocking of executables installed in the system


xasopheno:
In case the script uses commands that are not installed on the system that the tests are run.
``` yaml
dependencies:
#       this can be install here: asdf.com
        - gcsfuse
protocols:
        - protocol: []
```

soenkehahn:
Fixed through #187.

soenkehahn:
I reverted #187, since it didn't fully work. The problem is that it didn't work on machines where executables where installed in non-standard locations, e.g. `$HOME/.cargo/bin`. For an example, see https://github.com/Originate/scriptkeeper/tree/sh/example.

PR #177, opened 2019-03-25
add better example for heroku deploy script

soenkehahn:


tests/examples/heroku_deploy/script.protocols.yaml
```diff
@@ -0,0 +1,45 @@
...
+      PATH: /snap/bin:~/.cargo/bin:/usr/bin:/bin
+    arguments: test-heroku-app
+    protocol:
+      - command: git status --porcelain
+        stdout: " M some-changed-file\n"
+    stderr: "exiting, git repo not clean\n"
+    exitcode: 1
+
+  # exits and prints a usage message when no arguments given
+  - env: *env
```
soenkehahn: @matthandlersux: Look at those fancy yaml features that just work. :)

matthandlersux: it's pretty nice that we get that for free with the yaml parsing library

soenkehahn:
Sounds like a good idea.

soenkehahn:
@hallettj: I was thinking we could implement #114 before merging this. Then it would work on all machines. (It's also failing on CI right now, since `heroku` is not installed there.)

data-pup:
++ to the `examples/` point made by @matthandlersux above. The `examples/` is also special in that it is automatically compiled when you run your tests, so that they don't go out of date.


https://doc.rust-lang.org/cargo/reference/manifest.html#examples

PR #179, opened 2019-03-25
move file mocking tests into their own file

soenkehahn:


tests/files.rs
```diff
@@ -0,0 +1,72 @@
...
+mod utils;
+
+use scriptkeeper::R;
+use utils::test_run;
+
+#[test]
+fn allows_to_mock_files_existence() -> R<()> {
+    test_run(
+        r"
+                |#!/usr/bin/env bash
```
matthandlersux: looks like the indentation of these strings got messed up in the move

soenkehahn: Good catch.

PR #184, opened 2019-04-01
rename 'protocol' to 'steps' and 'protocols' to 'tests'

### LINKED ISSUE #182: Rename yaml fields
I'm wondering whether we should rename `protocols` to `tests` and `protocol` to `steps`. This might be easier to understand without learning the term 'Protocol' that only `scriptkeeper` uses.

This would be a breaking change, but I think that's fine at this point.

soenkehahn:
Closes #182.


README.md
```diff
@@ -132,87 +135,87 @@ unmockedCommands: [string]
...
       - git pull
 ```
 
-## Recording protocols
+## Recording tests
 
-There is **experimental** support for recording protocols. You can either record
-protocols by passing in the `--record` command line flag, or you can put
-so-called holes into your protocols:
+There is **experimental** support for recording tests. You can either record
```
matthandlersux: it just occurred to me that this idea might need a bit more explanation. i don't believe there are other similar tools out there that can automatically generate tests for you, so the idea of "recording tests" isn't obvious IMO.

soenkehahn: Yeah, that's right. This could probably use recording snapshot tests as an analogy.

Do you think that this PR makes the readme worse? If not I would defer your concern to a different issue/PR.

matthandlersux: no i think the change is fine, but using a more common word made it clearer to me that this was confusing. yes, i would defer to a separate PR. i can make a ticket for it.

PR #187, opened 2019-04-03
allow to test scripts that execute commands that don't exist on the s…

soenkehahn:
…ystem running the test

This is a solution for #114, at least in most situations. I went with a different approach than sketched out in that ticket: `scriptkeeper` will pretend that *all* executables exist in `/bin/`. So if a bash script executes a non existing command `foo`, it'll just execute as if that file existed.

@matthandlersux: Since you're using `scriptkeeper` with some mocked files and potentially uninstalled executables, do you see any problem with this?


src/test_spec/executable_path.rs
```diff
@@ -98,4 +112,33 @@ mod canonicalize {
...
+        }
+    }
+}
+
+pub fn is_in_bin(executable: &Path) -> bool {
+    let file_name = match executable.file_name() {
+        None => return false,
+        Some(f) => f,
+    };
+    executable.starts_with("/bin/") && PathBuf::from("/bin/").join(file_name) == executable
```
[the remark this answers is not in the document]
soenkehahn: Yeah, I think we don't need to worry about this. `scriptkeeper` only works on Linux for now and even if we manage to get it to work on macs, we're still fine with forward slashes.

soenkehahn:
> i think this is fine, but from an OSX perspective (and with the dependency on /bin/) i think it will need an update in the future if #137 lands. on OSX it's pretty common for any additionally added executables to end up in /usr/local/bin (because of homebrew, the popular package manager).

I don't think this conflicts with executables installed in `/usr/local/bin`. On Linux, binaries end up in `/bin`, `/usr/bin` and `/usr/local/bin` (among others). For the feature of this PR, we just have to pick one directory that hopefully is in people's `$PATH`. `/bin` seemed like a good choice.

src/executable_path.rs
```diff
@@ -98,4 +112,44 @@ mod canonicalize {
...
+        fn does_not_shorten_executables_in_subdirectory_of_bin_that_do_not_exist() -> R<()> {
+            let executable = Path::new("/bin/sub/doesnotexist");
+            let file_name = canonicalize(executable);
+            assert_eq!(file_name, PathBuf::from("/bin/sub/doesnotexist"));
+            Ok(())
+        }
+    }
+}
+
+pub fn is_in_bin(executable: &Path) -> bool {
```
soenkehahn: Rename to something that mentions that this is for pretending that executables exist. `is_fake_executable`? `is_pretend_executable`?

hallettj: How about `should_assume_in_path` ?

soenkehahn: Yes, thanks!

PR #189, opened 2019-04-04
better errors in case the scriptkeeper_executable can't be found

soenkehahn:
This can happen during tests when running `cargo test --lib` when
the scriptkeeper executable wasn't built before. The proper fix
would be to actually make sure we build the executable before
running these tests, but I'm not sure that's possible. In any
case this error message makes it easier to track down.

### DIFF
--- a/src/test_checker/executable_mock.rs
+++ b/src/test_checker/executable_mock.rs
@@ -13,6 +13,12 @@ pub struct Config {
 }
 
 pub fn create_mock_executable(context: &Context, config: Config) -> R<Vec<u8>> {
+    if !context.scriptkeeper_executable().exists() {
+        Err(format!(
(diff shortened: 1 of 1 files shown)


src/test_checker/executable_mock.rs
```diff
@@ -74,4 +81,21 @@ mod test {
         assert_eq!(output.status.code(), Some(42));
         Ok(())
     }
+
+    #[test]
+    fn aborts_with_a_helpful_message_when_scriptkeeper_executable_does_not_exist() {
```
xasopheno: clear as can be. :)

soenkehahn:
Yeah, I didn't want to mention how to fix this for tests in the error message, because this code is also executed in production. I don't know under which circumstances this can ever be triggered in production, but I didn't want to confuse anyone who might run into this in the future.

PR #190, opened 2019-04-04
treat commands as unmocked, even if they're not in the path

soenkehahn:
I think this is much more consistent, especially since the `PATH` value that we used to figure out, whether something is unmocked is not necessarily the `PATH` value that the script uses to resolve executables.


scriptkeeper-in-docker.sh.test.yaml
```diff
@@ -1,5 +1,5 @@
 unmockedCommands:
-  - /usr/bin/basename
+  - basename
```
soenkehahn: I noticed that you can't specify `unmockedCommands` as absolute paths anymore after this change. I think this test file was written when the executable shortform wasn't supported yet. So I still think this is a good change. @hallettj

soenkehahn:
After discussions with @hallettj, closing this. This should all be dealt with differently in #187.

PR #191, opened 2019-04-05
allow to specify expected stdout in test

### LINKED ISSUE #101: allow to specify the expected stdout of a tested script

hallettj:
fixes #101 

This PR also changes the `test_run` family of test helpers to take a value of a new type, `Expect`, which provides more flexibility for specifying expected results.

### DIFF
--- a/src/test_checker/mod.rs
+++ b/src/test_checker/mod.rs
@@ -4,7 +4,7 @@ pub mod executable_mock;
 use crate::context::Context;
 use crate::test_spec;
 use crate::test_spec::Test;
-use crate::tracer::stdio_redirecting::Redirector;
+use crate::tracer::stdio_redirecting::{Redirect, Redirector};
 use crate::tracer::{tracee_memory, SyscallMock};
 use crate::utils::short_temp_files::ShortTempFile;
 use crate::R;
@@ -69,6 +69,28 @@ impl TestChecker {
         let path = temp_executable.path();
         self.temporary_executables.push(temp_executable);
         Ok(path)
+    }
+
+    fn check_expected_output_stream(&mut self, redirect: &Redirect, expected: Vec<u8>) -> R<()> {
+        match redirect.captured()? {
+            None => panic!(
+                "scriptkeeper bug: {} expected, but not captured",
+                redirect.stream_type
+            ),
+            Some(captured) => {
+                if captured != expected {
+                    self.result.register_error(format!(
+                        "  expected output to {}: {:?}\
+                         \n  received output to {}: {:?}\n",
+                        redirect.stream_type,
+                        String::from_utf8_lossy(&expected).as_ref(),
+                        redirect.stream_type,
+                        String::from_utf8_lossy(&captured).as_ref(),
+                    ));
+                }
+            }
+        }
+        Ok(())
     }
 }
 
@@ -148,20 +170,11 @@ impl SyscallMock for TestChecker {
                 &format!("<exitcode {}>", exitcode),
             );
         }
+        if let Some(expected_stdout) = &self.test.stdout {
+            self.check_expected_output_stream(&redirector.stdout, expected_stdout.clone())?;
+        }
         if let Some(expected_stderr) = &self.test.stderr {
-            match redirector.stderr.captured()? {
-                None => panic!("scriptkeeper bug: stderr expected, but not captured"),
-                Some(captured_stderr) => {
-                    if &captured_stderr != expected_stderr {
-                        self.result.register_error(format!(
-                            "  expected output to stderr: {:?}\
-                             \n  received output to stderr: {:?}\n",
-                            String::from_utf8_lossy(&expected_stderr).as_ref(),
-                            String::from_utf8_lossy(&captured_stderr).as_ref(),
-                        ));
-                    }
-                }
-            }
+            self.check_expected_output_stream(&redirector.stderr, expected_stderr.clone())?;
         }
         Ok(self.result)
     }
--- a/src/test_checker/mod.rs
+++ b/src/test_checker/mod.rs
@@ -15,6 +15,7 @@ use nix::unistd::Pid;
 use std::ffi::OsString;
 use std::os::unix::ffi::OsStrExt;
 use std::path::PathBuf;
+use trim_margin::MarginTrimmable;
 
 #[derive(Debug)]
 pub struct TestChecker {
@@ -79,14 +80,21 @@ impl TestChecker {
             ),
             Some(captured) => {
                 if captured != expected {
-                    self.result.register_error(format!(
-                        "  expected output to {}: {:?}\
-                         \n  received output to {}: {:?}\n",
-                        redirect.stream_type,
-                        String::from_utf8_lossy(&expected).as_ref(),
-                        redirect.stream_type,
-                        String::from_utf8_lossy(&captured).as_ref(),
-                    ));
+                    self.result.register_error(
+                        format!(
+                            r"
+                                |  expected output to {}: {:?}
+                                |  received output to {}: {:?}
+                                |
+                            ",
+                            redirect.stream_type,
+                            String::from_utf8_lossy(&expected).as_ref(),
+                            redirect.stream_type,
+                            String::from_utf8_lossy(&captured).as_ref(),
+                        )
+                        .trim_margin()
+                        .unwrap(),
+                    );
                 }
             }
         }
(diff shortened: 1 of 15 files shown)


soenkehahn:
Could you squash all the changes related to `Expect` into one commit? I think that would make it really easy to review this PR.

hallettj:
@soenkehahn Sure thing - I squashed and reordered some commits.

tests/utils.rs
```diff
@@ -46,15 +81,22 @@ pub fn test_run_with_context(
     context: &Context,
```
soenkehahn: Do we still need `test_run_with_context` to be public? Or can we replace that with `test_run` everywhere, new that matching results is smarter?

tests/utils.rs
```diff
@@ -14,12 +14,47 @@ use std::path::PathBuf;
...
 use yaml_rust::YamlLoader;
 
-fn compare_results(result: (ExitCode, String), expected: Result<(), &str>) {
-    let expected_output = match expected {
-        Err(expected_output) => (ExitCode(1), expected_output.to_string()),
-        Ok(()) => (ExitCode(0), "All tests passed.\n".to_string()),
-    };
-    assert_eq!(result, expected_output);
+#[derive(Debug, PartialEq)]
+pub struct Expect {
```
matthandlersux: "Expect" feels a bit generic for what this does. the verbose description i can think of is "ScriptRunEndState", but that's probably too wordy. any ideas?

tests/utils.rs
```diff
@@ -14,12 +14,47 @@ use std::path::PathBuf;
...
+
+    pub fn err(expected_output: &str) -> Self {
+        Expect {
+            expected_exit_code: ExitCode(1),
+            expected_stdout: expected_output.to_string(),
+            expected_stderr: "".to_string(),
+        }
+    }
+
+    pub fn stdout(self, expected_output: &str) -> Self {
```
matthandlersux: since the usage is `Expect::ok().stdout(...)` my vote would be to name this in a way that suggests you're adding to the structure rather than performing some action, something like `Expect::ok().with_stdout(...)`.

src/test_checker/mod.rs
```diff
@@ -70,6 +70,28 @@ impl TestChecker {
...
+    fn check_expected_output_stream(&mut self, redirect: &Redirect, expected: Vec<u8>) -> R<()> {
+        match redirect.captured()? {
+            None => panic!(
+                "scriptkeeper bug: {} expected, but not captured",
+                redirect.stream_type
+            ),
+            Some(captured) => {
+                if captured != expected {
+                    self.register_error(format!(
+                        "  expected output to {}: {:?}\
```
matthandlersux: i realize this was moved from below, but the `trim_margin` function is nice if we can use it here, i think it makes things much more readable

tests/utils.rs
```diff
@@ -14,12 +14,47 @@ use std::path::PathBuf;
...
+        Expect {
+            expected_stderr: expected_output.to_string(),
+            ..self
+        }
+    }
+}
+
+fn compare_results(actual: Expect, expected: Expect) {
+    assert_eq!(actual, expected);
 }
```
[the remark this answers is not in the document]
hallettj: Good point, inlined.

tests/utils.rs
hallettj: Yes, and we can also inline `test_run_with_context`.

src/tracer/stdio_redirecting.rs
```diff
@@ -47,44 +45,41 @@ impl Redirector {
...
     write_end: RawFd,
     captured: Option<Arc<Mutex<Cursor<Vec<u8>>>>>,
 }
 
 impl Redirect {
-    fn new(context: &Context, stream_type: StreamType) -> R<Redirect> {
-        Redirect::new_internal(context, stream_type, None)
-    }
-
-    fn new_capturing(context: &Context, stream_type: StreamType) -> R<Redirect> {
```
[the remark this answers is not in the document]
hallettj: Ok, I split out those two constructor variants.

tests/stdio.rs
```diff
@@ -49,12 +39,94 @@ fn relays_stderr_from_the_tested_script_to_the_user() -> R<()> {
...
+                    |  expected output to stdout: "foo\n"
+                    |  received output to stdout: ""
+                "#,
+            )?),
+        )?;
+        Ok(())
+    }
+
+    #[test]
+    fn when_not_specified_but_scripts_writes_to_stdout() -> R<()> {
```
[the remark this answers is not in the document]
hallettj: Done

tests/stdio.rs
```diff
@@ -119,4 +192,20 @@ mod expected_stderr {
         )?;
         Ok(())
     }
+
+    #[test]
+    fn when_not_specified_but_scripts_writes_to_stderr() -> R<()> {
```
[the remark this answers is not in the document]
hallettj: I reworded this test name and the name of the corresponding stdout test a little bit. The detail we are testing is that scriptkeeper tests pass when there is output if no expected output has been specified.

tests/utils.rs
hallettj: My preference is to put descriptiveness in the method names, as opposed to the name of the type itself. To that end I changed the names of the `Expect` methods:

- `ok` → `tests_pass`
- `err` → `error_message`
- `stdout` → `with_stdout` (per your suggestion)
- `stderr` → `with_stderr`

Although this type is specialized for our domain, I think it would be nice for it to read like a sentence as expectations in some test frameworks do. For example, `Expect::error_message("failure").with_stdout("test")`

tests/utils.rs
hallettj: Good idea! I have made that change.

src/test_checker/mod.rs
hallettj: Sounds good. I moved the trim-margin dependency from the build dependencies list to general dependencies to make this work.

tests/run.rs
```diff
@@ -198,7 +198,7 @@ fn failing() -> R<()> {
             |steps:
             |  - cp
         ",
-        Err(&trim_margin(
+        Expect::error_message(&trim_margin(
```
soenkehahn: Do we want to move the call to `trim_margin` into `Expect::error_message`? Do we ever use `error_message` without it?

tests/utils.rs
```diff
@@ -42,22 +73,21 @@ pub fn test_run_with_tempfile(
     Ok((exitcode, context.get_captured_stdout()))
 }
```
soenkehahn: Could `test_run_with_tempfile` now be changed to not return `(ExitCode, String)` and instead take an argument of type `Expect`?

src/test_checker/mod.rs
```diff
@@ -70,6 +71,35 @@ impl TestChecker {
...
+                    );
+                }
+            }
+        }
+        Ok(())
+    }
+
     fn register_step_error(&mut self, expected: &str, received: &str) {
         self.register_error(format!(
             "  expected: {}\n  received: {}\n",
```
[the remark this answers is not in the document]
hallettj: Sure thing, I've made that change.

tests/run.rs
hallettj: You're right, we use `trim_margin` in every case where we use `Expect::error_message` I made this change; and now `Expect::error_message` returns `R<Expect>` instead of `Expect`.

tests/utils.rs
hallettj: As we discussed passing an `Expect` to `test_run_with_tempfile` does not work out. But I simplified `test_run_with_tempfile` to return just an exit code.

hallettj:
@soenkehahn How do things look after my latest set of changes?

matthandlersux:
@soenkehahn i believe this PR is just waiting on your final review

soenkehahn:
Yeah, sorry. I think finishing up the first binary release has a higher priority right now. I hope I'll look at this soon.

soenkehahn:
@hallettj: I took the liberty to merge more commits and rebase on current master. Mostly I combined later refactorings into commits that introduced the code change in the first place. I think I'd like to have the introduction of `Expect` and the support for expected `stdout` as separate commits on master, so I didn't want to squash this PR. But then I also didn't want to have all these small refactoring commits you had before.

It's also much easier to review now. And LGTM. Should I merge or do you still have any comments/changes?

hallettj:
Sounds good. If it looks good to you then please feel free to merge.

soenkehahn:
@hallettj: Thanks a lot for the PR. :)

PR #194, opened 2019-04-08
allow to mock stderr output of test step commands

hallettj:
This change is branched from `sh/expected-output` (pr #192) because I wanted to use the `expect_bytes` method that I added in that branch.


soenkehahn:
@hallettj: Do you want to rebase this branch against current master? Or should I take over this PR?

hallettj:
Either way works for me. I will rebase next week if you don't get to it
before then.

PR #196, opened 2019-04-09
add test pattern parameter to `just dev`

soenkehahn:
This allows us to do `just dev some_test_to_run` (in addition to `just dev`).


soenkehahn:
@matthandlersux: Added something to the readme.

PR #197, opened 2019-04-17
rustc and clippy version update

soenkehahn:
I added `allow(clippy::redundant_closure)` in two places. Without that we get a lot of clippy errors like this:
```
  --> src/test_checker/checker_result.rs:55:27
   |
55 |         self.0.iter().all(|result| result.is_pass())
   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^ help: remove closure as shown: `test_checker::checker_result::CheckerResult::is_pass`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
```

I think these are nicer readable with that `.`-method-syntax.


soenkehahn:
@matthandlersux: Hmm, I don't know. Here's all the `clippy` rules: https://rust-lang.github.io/rust-clippy/master/index.html. The one that is disabled in this PR is this: https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure.

PR #199, opened 2019-04-24
add script to build a distributable binary

soenkehahn:
This PR adds a script that allows to build a binary of scriptkeeper into `distribution/scriptkeeper`. It's built inside docker with an older centOS image, so that it doesn't depend on a newer version of glibc. This will hopefully make it work on more distributions. There's also a test script that makes sure that the dynamic dependencies can be found on a bunch of distributions.

My plan is to merge this and then cut a release (I think version `0.2.0`) and upload that binary manually to the release on github. (We can also automate that process later.)

### DIFF
--- a/distribution/Dockerfile
+++ b/distribution/Dockerfile
@@ -1,0 +1,18 @@
+FROM centos:6.10
+RUN yum install --assumeyes gcc
+
+# install rust toolchain
+WORKDIR /root
+RUN curl https://sh.rustup.rs -sSf >> rustup.sh
+RUN chmod +x rustup.sh
+RUN ./rustup.sh -y
+ENV PATH=/root/.cargo/bin:$PATH
+
+# build scriptkeeper
+WORKDIR /root/scriptkeeper
+ADD Cargo.* ./
+RUN mkdir src && touch src/lib.rs && cargo install --root /usr/local --path . ; true
+ADD src ./src
+RUN touch src/lib.rs
+RUN cargo install --root /usr/local --path .
+RUN strip /usr/local/bin/scriptkeeper
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /target/
+/distribution/scriptkeeper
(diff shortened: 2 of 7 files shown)


distribution/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM centos:6.10
+RUN yum install --assumeyes gcc
```
matthandlersux: there's a funny bit in the latest season of arrested development that this reminds me of

presume teeth

matthandlersux: there's a funny bit in the latest season of arrested development that this reminds me of

presume teeth

soenkehahn:
> update the readme to describe your intended release process (and that way others can follow it too)

I added something to the readme. I'm not 100% sure this is how everything will work, since I've never done it, but we can update that section later.

> should we write tests for build script?

Done.

> should we add a step to CI that runs test-distribution.sh on a certain branch (i'm not sure if circle lets you interact directly with docker from within a container)? maybe we update the default branch on this repo to be something else, and save master for releases?

Hmm, I didn't even plan to necessarily run this on every release. This is more a one-time script to make sure the dynamically linked dependencies work out. So as long as we don't change how the distributed binary is built, we don't need to re-run this. Is there a good way to make this more obvious? Maybe rename the script to `distribution/test-dynamic-dependencies.sh`? And maybe leave a comment in the script?

matthandlersux:
> Hmm, I didn't even plan to necessarily run this on every release. This is more a one-time script to make sure the dynamically linked dependencies work out. So as long as we don't change how the distributed binary is built, we don't need to re-run this. Is there a good way to make this more obvious? Maybe rename the script to distribution/test-dynamic-dependencies.sh? And maybe leave a comment in the script?

i think a comment in the readme in the section about cutting a new release would make sense. if the comment is in the script itself, then people will only discover it if they go searching.

WRT the tests for the build script... should those run on CI?

soenkehahn:
I added a comment about `distribution/test-distribution.sh` to the readme and `distribution/build.sh` is now tested on Ci.

PR #204, opened 2019-05-30
pin down rustc version in both Dockerfiles

soenkehahn:
This PR pins down the rustc version (to `1.34.2`) for 

- `distribution/Dockerfile` and
- `./Dockerfile`, which is used for running and developing `scriptkeeper` on mac.

The motivation for this PR was to pin down the compiler version for the distributed executable. But I thought it'd be good to DRY up setting the compiler version between the two Dockerfiles. In it's current form that does mean that you have to build the docker image on macs by saying `just build_docker_image` instead of calling `./build-docker-image.sh`.

@matthandlersux: I think you're the only one running this on mac currently, any opinions on this?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM rust:1.34.0
+ARG RUSTC_VERSION
+FROM rust:${RUSTC_VERSION}
 
 RUN cargo install cargo-watch
 RUN apt-get update && apt-get install -y ruby
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: rust:1.34.0-stretch
+      - image: rust:1.34.2-stretch
 
     steps:
       - run:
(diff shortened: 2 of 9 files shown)


Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM rust:1.34.0
+ARG RUSTC_VERSION
+FROM rust:${RUSTC_VERSION}
 
 RUN cargo install cargo-watch
 RUN apt-get update && apt-get install -y ruby
```
hlian: Unrelated to this PR, and probably would fit better in another PR: This Dockerfile is building as root, which [isn't recommended](http://canihaznonprivilegedcontainers.info). You can do `useradd --create-home myusernamehere` followed by `USER myusernamehere` to drop privileges; [here's an example](https://pythonspeed.com/articles/dockerizing-python-is-hard/).

soenkehahn: Hmm, building as non-root is a bit ugly, because of this issue: https://github.com/moby/moby/issues/6119. Building as root and than just **running** the executable as non-root would be easier. What exactly are the security concerns?

matthandlersux:
i think you'll need to update the mac section in the readme, though. it still points at `./build-docker-image.sh`

soenkehahn:
@matthandlersux: Yes, good catch.

PR #205, opened 2019-05-30
move register_error and register_step_error into CheckerResult

### LINKED ISSUE #201: fix compilation for rustc-1.35.0
There's a new warning popping up for the new rust compiler when running `just ci`.

soenkehahn:
Fixes #201.

### DIFF
--- a/src/test_checker/checker_result.rs
+++ b/src/test_checker/checker_result.rs
@@ -28,6 +28,22 @@ impl CheckerResult {
             },
         }
     }
+
+    pub fn register_step_error(&mut self, expected: &str, received: &str) {
+        self.register_error(format!(
+            "  expected: {}\n  received: {}\n",
+            expected, received
+        ));
+    }
+
+    pub fn register_error(&mut self, message: String) {
+        match self {
+            CheckerResult::Pass => {
+                *self = CheckerResult::Failure(message);
+            }
+            CheckerResult::Failure(_) => {}
+        }
+    }
 }
 
 pub struct CheckerResults(pub Vec<CheckerResult>);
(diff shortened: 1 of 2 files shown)


src/test_checker/mod.rs
```diff
@@ -168,7 +153,7 @@ impl SyscallMock for TestChecker {
                 None => panic!("scriptkeeper bug: stderr expected, but not captured"),
                 Some(captured_stderr) => {
                     if &captured_stderr != expected_stderr {
-                        self.register_error(format!(
+                        self.result.register_error(format!(
```
soenkehahn: This is the place where #201 is actually fixed. `rustc-1.35.0` warned about self being mutably borrowed here (by `register_error`), but also immutably borrowed 5 lines above in `&self.test.stderr`. Since `register_error` was now moved from `TestChecker` to `CheckerResult`, this no longer clashes.

An alternative fix would be to clone `stderr` above. @matthandlersux, @data-pup: Any opinions or other ideas?

src/test_checker/checker_result.rs
```diff
@@ -28,6 +28,22 @@ impl CheckerResult {
...
     }
+
+    pub fn register_step_error(&mut self, expected: &str, received: &str) {
+        self.register_error(format!(
+            "  expected: {}\n  received: {}\n",
+            expected, received
+        ));
+    }
+
+    pub fn register_error(&mut self, message: String) {
```
hlian: not &str here?

soenkehahn: We store that string in `CheckerResult::Failure` here. And I don't want to make that field `&str`, since then that type would require lifetime parameters.

src/test_checker/mod.rs
matthandlersux: no opinion, this seems to work just fine

PR #206, opened 2019-05-30
run scriptkeeper tests for distribution/build.sh on ci

soenkehahn:
Maybe we should try to use `just` on CI to have only one single source of truth for how to run the tests. Anyways, this PR brings what's in the `Justfile` and `.circleci/config.yml` closer together.


matthandlersux:
running `just` on CI makes sense to me, it's nice to avoid as much duplication as possible on the CI side. if we go that route, might as well ensure that `just` is used when running in docker on a mac, as well.

PR #212, opened 2019-09-21
fix multiple tests example in README.md

mahermalaeb:
To my understanding, the `arguments` parameter should be `pull` instead of push `push` when we expect the `git pull` to be executed. This will also avoid having different tests with same arguments.


soenkehahn:
Yeah, that's a typo. Thanks!

soenkehahn (approved):

